PR #13, opened 2023-03-14
`POST /capacity` + `GET /capacity` + `GET /capacity/{id}`

jonmattgray:
- Adds `POST /capacity` + `GET /capacity` + `GET /capacity/{id}`
image

- @decorators for OpenAPI descriptions
- A blank `authenticationModule` to pass through `Authentication` headers for external validation
image

- Integration tests with seeds and mocked identity service


docker-compose.yml
```diff
@@ -2,15 +2,49 @@
 version: '3'
 
 services:
-  db:
+  postgres-matchmaker-api:
     image: postgres:15.0-alpine
```
dblane-digicatapult, line 6: Upgrade postgres container version

package.json
```diff
@@ -55,7 +55,8 @@
     "prettier": "^2.8.4",
     "supertest": "^6.3.3",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.5"
+    "typescript": "^4.6.4",
```
dblane-digicatapult, line 58: Why the downgrade?

jonmattgray: whoops, I messed up merging the package.json with main, probably from that

docker-compose.yml
jonmattgray: bumped to `15.2`

test/helper/identityMock.ts
```diff
@@ -0,0 +1,40 @@
+import { MockAgent, setGlobalDispatcher } from 'undici'
```
n3op2, line 1: mmm?? do we need to use this? If we are intercepting and etc, then stub maybe?

However, can see quite a few use cases of this, myself I used nock so would be great to get Matts opinion on this since we discussed something about builtin `fetch`

test/helper/routeHelper.ts
```diff
@@ -0,0 +1,25 @@
+import request from 'supertest'
+import express from 'express'
+
+export async function get(app: express.Express, endpoint: string) {
```
n3op2, line 4: why this is async? it returns a promise so no need .then/.catch and does not perform async within a function, can we just return?

We can return promise now?
```ts
export get = (...) => request(...
```

src/lib/db/index.ts
```diff
@@ -21,13 +23,35 @@ export default class Database {
...
+      .select(['id', 'owner', 'status', 'parameters_attachment_id AS parametersAttachmentId'])
+      .where({ subtype })
+  }
+
+  getDemand = async (capacityId: UUID, subtype: DemandSubtype) => {
+    return this.db()
+      .demands()
+      .select(['id', 'owner', 'status', 'parameters_attachment_id AS parametersAttachmentId'])
+      .where({ id: capacityId, subtype })
+  }
```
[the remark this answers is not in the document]
jonmattgray: how would we handle the `select`s and `where`s with that?

src/lib/error-handler/index.ts
```diff
@@ -7,6 +7,29 @@ export interface ValidateErrorJSON {
...
+  constructor({ code = 500, message = 'Internal server error' }) {
+    super(message)
+    this.code = code
+    this.message = message
+  }
+}
+
+export class BadRequestError extends HttpResponseError {
+  constructor(message: string) {
+    super({ code: 400, message: `Bad Request: ${message}` })
```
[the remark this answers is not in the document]
jonmattgray: where is it doubling up? `code` gets picked up by the error handler to build a response:
```
  if (err instanceof HttpResponseError) {
    logger.debug(`Bad request for ${req.path}`)
    return res.status(err.code).json(err.message)
  }
```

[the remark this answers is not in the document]
jonmattgray: ah I've seen your other comment about using `@Response<Type>(status)` instead

test/helper/identityMock.ts
jonmattgray: `nock` doesn't work with node 18 fetch

jonmattgray: there will be more elaborate intercepting when we get to testing `accept_a` or `accept_b` first flows

test/helper/routeHelper.ts
jonmattgray: updated to tidy them up

src/services/identity.ts
```diff
@@ -0,0 +1,34 @@
...
+    const member = await res.json()
+    return member.address
+  }
+
+  throw new HttpResponseError({})
+}
+
+const getMemberByAddress = (alias: string) => getMemberByAlias(alias)
+
+export { getMemberByAlias, getMemberByAddress, getMemberBySelf }
```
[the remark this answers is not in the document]
jonmattgray: moved

src/lib/error-handler/index.ts
```diff
@@ -7,6 +7,29 @@ export interface ValidateErrorJSON {
   details: { [name: string]: unknown }
 }
 
+export class HttpResponseError extends Error {
```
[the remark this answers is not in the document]
jonmattgray: added
image

src/lib/error-handler/index.ts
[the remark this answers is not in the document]
jonmattgray: I think the `code` property might be necessary so the error handler uses it to respond with the correct code. I'm not sure there's a way of passing the e.g. `400` in `  @Response<BadRequestError>(400, 'Request was invalid')` automatically through without catching it in a handler, have you found a way?

src/controllers/capacity/index.ts
```diff
@@ -0,0 +1,93 @@
...
+   */
+  @Get('/')
+  public async getAll(): Promise<DemandResponse[]> {
+    const capacities = await this.db.getDemands(DemandSubtype.Capacity)
+    const result = await Promise.all(capacities.map(async (capacity: DemandResponse) => responseWithAlias(capacity)))
+    return result
+  }
+
+  /**
+   * @summary Get a capacity by ID
```
n3op2, line 70: nice, going to add these as well:)

src/controllers/capacity/index.ts
```diff
@@ -0,0 +1,93 @@
+import { Controller, Get, Post, Route, Path, Response, Body, SuccessResponse, Tags, Security } from 'tsoa'
+import { Logger } from 'pino'
+
+import { logger } from '../../lib/logger'
+import Database from '../../lib/db'
+import { DemandResponse, DemandSubtype, DemandRequest, DemandStatus } from '../../models/demand'
+import { UUID } from '../../models/uuid'
+import { NotFoundError } from '../../lib/error-handler/index'
```
n3op2, line 8: do we nee to suffix with Error, they are already in lib/error-handler (maybe errors would be more appropriate)

src/lib/error-handler/index.ts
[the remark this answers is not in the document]
n3op2: I think I did if you look at the PR you will find. Ended up using code property so it's easier for use to sync up.

```ts
  if (err instanceof Error) { // probly in your case HttpError liking this wrapper, could have status/code and then check if instance of HttpErrr and return with code if not 500 + err 
    return res.status(err.code || 500).json(err)
  }
```

src/controllers/capacity/index.ts
jonmattgray: removed them

tsoa.json
```diff
@@ -4,9 +4,17 @@
...
+        "type": "http",
+        "scheme": "bearer",
+        "bearerFormat": "JWT"
+      }
+    }
   },
   "routes": {
-    "routesDir": "src"
+    "routesDir": "src",
+    "authenticationModule": "./src/authentication.ts"
```
n3op2, line 18: Mmm... since it's a module maybe lets put it in lib/? since it's around 10 lines of code maybe include into `server.ts` and then when we have an x number of middleware we can break down. either way... if possible without hustle would prefer it to not be at root of `src/` but again this is not stopping me from approving.

n3op2 (approved):

src/lib/error-handler/index.ts
```diff
@@ -21,7 +44,10 @@ export const errorHandler = function errorHandler(
     }
     return res.status(422).json(response)
   }
-  if (err instanceof Error) {
+  if (err.code) {
```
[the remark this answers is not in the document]
jonmattgray: changed back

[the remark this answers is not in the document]
jonmattgray: changed back

tsoa.json
jonmattgray: moved to `/lib`

PR #15, opened 2023-03-15
Feature/l3 25

n3op2:
# What

### changes
- new model - attachments
- a few endpoints [list/get/create]
- a few little changes to types - better defined. however, one ANY left but this will change with change to db lib
- new errors have been added [bad request, not found]

### swagger screenshots

Screenshot 2023-03-15 at 11 37 46
Screenshot 2023-03-15 at 11 37 39
Screenshot 2023-03-15 at 15 09 08


src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,120 @@
...
+  message?: string
+  data?: Attachments[] | Attachments
+  headers?: { [k: string]: unknown }
+}
+
+type File = {
+  [k: string]: string
+}
 
 @Route('attachments')
```
jonmattgray, line 25: based off our previous conventions, the route should be `attachment` rather than `attachments`?

src/lib/error-handler/index.ts
```diff
@@ -2,9 +2,51 @@ import { Response as ExResponse, Request as ExRequest, NextFunction } from 'expr
...
+    this.item = item
+    this.name = 'not found'
+    // this.stack = (<any> new Error()).stack
+  }
+}
+
+/**
+ * indicates that request was invalid e.g. missing parameter
+ */
+export class BadRequst extends Error implements IBadRequest {
```
jonmattgray, line 42: `BadRequst` -> `BadRequest`

src/controllers/attachments/index.ts
n3op2: mmm... happy to change, but it has to be inline with database model. Not sure why we have called plural. Since on row = one file? I might be wrong.

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,120 @@
...
+    this.log.debug(`crating an attachment ${JSON.stringify(file || req.body)}`)
+    if (file) {
+      await this.db.attachments().insert({
+        filename: file.originalname,
+        binary_blob: Buffer.from(file.buffer),
+      })
+      return {
+        status: 201,
+        message: 'created as file blob',
+      }
```
jonmattgray, line 63: do we need to include `status codes in the body?

n3op2: yes, TSOA then forward it. You can use `@SuccessResponse()` decorator so in this instance you can say

```ts
@SuccessResponse<Success>
.... (req: Request...
```

src/controllers/attachments/index.ts
jonmattgray: it doesn't need to be inline with the database model, I'm using `@Route('capacity')` with a `Demands` model

src/controllers/attachments/index.ts
jonmattgray: I think `@SuccessResponse` can be used without needing `status` in the body though? e.g. the following code returns a `201`

```
  @Post()
  @SuccessResponse('201')
  public async createCapacity(@Body() requestBody: DemandRequest): Promise<DemandResponse> {

   ...db logic

    return {
      id: capacity.id,
      owner: ownerAlias,
      status: capacity.status,
      parametersAttachmentId: parametersAttachmentId,
    }
  }
```

jonmattgray: then `response.body` doesn't contain `status` or `message` is can just be `'created as file blob'`

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,120 @@
-import { Controller, Get, Route } from 'tsoa'
+import { UploadedFile, Request, Controller, Get, Route, Path, Post, Header, Response } from 'tsoa'
 import { Logger } from 'pino'
+import express from 'express'
 
 import { logger } from '../../lib/logger'
-import Database from '../../lib/db'
+import Database, { Models, Query } from '../../lib/db'
 import type { Attachments } from '../../models'
+import { BadReqeust, NotFound } from '../../lib/error-handler'
```
jonmattgray, line 8: `BadReqeust` typo

n3op2: ha... I'm embarrassed. To be fair when I type I do trust my muscle memory. Sometimes I like how I replicate typo across without being aware.

src/controllers/attachments/index.ts
n3op2: Yes, indeed was doubling up on statuses. I've removed and unwrapped from data object since it's a body either way.

src/controllers/attachments/index.ts
n3op2: for tsoa yes, you can use any other route, but since the entity in database is **attachments** I would like to stick to it for obvious reasons. If database entity and route are totally different, like here

The whole idea is that TSOA model is inline with database model? Why because then we can utilize typescript for types and definitions, so no need to double up on definitions. At the moment is hard to see the benefit since the database is not fully implemented and we jsut create a connection

This will help us to have something called REST factory (get basic CRUD right of the model) e.g. would like to get template to that level so we just need to write code and use a strongly typed approach. To be fair easier would be to talk through

src/lib/error-handler/index.ts
```diff
@@ -2,9 +2,46 @@ import { Response as ExResponse, Request as ExRequest, NextFunction } from 'expr
...
 
 import { logger } from '../logger'
-export interface ValidateErrorJSON {
-  message: 'Validation failed'
-  details: { [name: string]: unknown }
+
+/**
+ * this should reflect database tables
+ */
+type Entities = 'attachments' // add as needed
```
jonmattgray, line 9: I'm not sure I see the purpose of having another type to manage here, could we not pass the entity as an argument to the error? Then it could be `item: string`

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,94 @@
...
-    this.db = this.dbClient.db()
+    // TMP will be updated with a wrapper so ORM client independant
+    this.db = this.dbClient.init()
   }
 
   @Get('/')
-  public async get(): Promise<{ status: number; attachments: Attachments[] }> {
-    this.log.debug({
-      msg: 'this is a test route to validate tsoa/swagger',
+  @SuccessResponse(201, 'returns all attachments')
```
jonmattgray, line 28: `200` rather than `201`

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,94 @@
...
+    const result = await this.db.attachments()
+
+    return result.map(({ binary_blob, ...item }: Attachments) => ({
+      ...item,
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachemnt has been created')
```
jonmattgray, line 40: `attachemnt` typo

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,94 @@
...
+
+    await this.db.attachments().insert({
+      filename: 'json',
+      binary_blob: Buffer.from(JSON.stringify(req.body)),
     })
 
-    return {
-      status: 200,
-      attachments: await this.db.attachments(),
+    return 'success'
```
jonmattgray, line 57: we surely want it to return the `id`? Otherwise you can't reference the attachment you just created in a subsequent API call

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,94 @@
...
+    return result.map(({ binary_blob, ...item }: Attachments) => ({
+      ...item,
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachemnt has been created')
+  public async create(@Request() req: express.Request, @UploadedFile() file: File): Promise<string> {
+    this.log.debug(`crating an attachment ${JSON.stringify(file || req.body)}`)
```
jonmattgray, line 42: `crating` typo

src/lib/error-handler/index.ts
```diff
@@ -14,18 +51,14 @@ export const errorHandler = function errorHandler(
...
+    return res.send(err)
   }
   if (err instanceof Error) {
-    logger.warn('Unexpected error thrown in handler: %s', err.message)
-    return res.status(500).json({
-      message: 'Internal Server Error',
-    })
+    logger.error('Unexpected error thrown in handler: %s', err.message)
+
+    return res.status(500).json(err)
```
jonmattgray, line 61: I think there needs to be code in the handler to handle each error type and return the correct code. At the moment the code is always `500` when there's an error. The `  @Response<BadRequest>(400)` type decorators don't seem to pass a code through?

jonmattgray: for example, getting an attachment with a bad id should `404` but it `500`s
image

src/controllers/attachments/index.ts
jonmattgray: the REST factory idea sounds good, but isn't implemented in this PR. And `@Route('attachments')` -> `@Route('attachment')` doesn't affect anything else. Our other APIs use single form rather than plural
https://github.com/digicatapult/morello-api/blob/cbefd92745b7d57288f935f98898f289e564f50f/src/controllers/scenario/index.ts#L10
image

src/controllers/attachments/index.ts
```diff
@@ -1,32 +1,94 @@
...
+import Database, { Models, Query } from '../../lib/db'
 import type { Attachments } from '../../models'
+import { BadReqeust, NotFound } from '../../lib/error-handler'
+
+type File = {
+  [k: string]: string
+}
 
 @Route('attachments')
 export class attachments extends Controller {
```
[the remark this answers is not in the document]
n3op2: This shouldn't be an issue any longer

src/lib/error-handler/index.ts
n3op2: This will keep us a little bit more inline with the records. String means that it can be anything in this instance this works like enums but without the negative side.

package.json
```diff
@@ -50,6 +50,7 @@
     "eslint-config-prettier": "^8.7.0",
     "eslint-plugin-prettier": "^4.2.1",
     "mocha": "^10.2.0",
+    "multer": "^1.4.5-lts.1",
```
mattdean-digicatapult, line 53: are we using `multer`?

src/lib/db/index.ts
```diff
@@ -10,31 +10,40 @@ import { UUID } from '../../models/uuid'
...
-      fs.readdirSync(MODELS_DIRECTORY).forEach((file: string): any => {
+    this.db = (models: Models<Query> = {}) => {
+      fs.readdirSync(MODELS_DIRECTORY).forEach((file: string) => {
         const { name } = path.parse(file)
         // TODO check if table exists -> append to the db object
-        if (name != 'index.d') models[name] = () => this.client(name)
+        if (name != 'index.d' && name != 'health') {
+          this.log.debug(`initializing ${name} db model`)
+          models[name] = () => this.client(name)
+        }
```
mattdean-digicatapult, line 39: This is pretty jank. It relies on the fact that no path transformations take place and if I'm right the only purpose is to get the name of the table? I'm not sure I've got a better solution right now but even a list of strings at the top of this file might be better with a runtime dynamic `import` call to ensure the file exists? At least that way if something like webpack were ever to run here it would see the dynamic import and know not to transform the names

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      this.setHeader('accept', 'application/octet-stream')
+      this.setHeader('access-control-expose-headers', 'content-disposition')
+      this.setHeader('content-disposition', `attachment; filename="${attachment.filename}"`)
+      this.setHeader('maxAge', `${365 * 24 * 60 * 60 * 1000}`)
+      this.setHeader('immutable', 'true')
+
+      return attachment
+    }
+
+    // no need to check for filename since JSON.parse will throw
```
mattdean-digicatapult, line 86: If `JSON.parse` throws what is returned to the use?

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    // TMP doubling upping on headers as I was not able to figure out why 'accept' using @Header() get overwritten
+    // also not checking for mime type as we have two options here download or sned json
+    if (type === 'file' || accept === 'application/octet-stream') {
+      this.setHeader('accept', 'application/octet-stream')
+      this.setHeader('access-control-expose-headers', 'content-disposition')
+      this.setHeader('content-disposition', `attachment; filename="${attachment.filename}"`)
+      this.setHeader('maxAge', `${365 * 24 * 60 * 60 * 1000}`)
+      this.setHeader('immutable', 'true')
+
+      return attachment
```
mattdean-digicatapult, line 83: This returns the whole `Attachment` right? This should only return the blob.

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    return result.map(({ binary_blob, ...item }: Attachment) => ({
+      ...item,
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachment has been created')
+  public async create(@Request() req: express.Request, @UploadedFile() file: File): Promise<string> {
+    this.log.debug(`crating an attachment ${JSON.stringify(file || req.body)}`)
```
mattdean-digicatapult, line 42: typeo on `creating`

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    return result.map(({ binary_blob, ...item }: Attachment) => ({
+      ...item,
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachment has been created')
+  public async create(@Request() req: express.Request, @UploadedFile() file: File): Promise<string> {
```
mattdean-digicatapult, line 41: Interesting so does this magically handle that it might be either a multi part form or a json body?

mattdean-digicatapult (commented):
Some comments to improve my understanding

src/lib/error-handler/index.ts
```diff
@@ -37,22 +61,22 @@ export const errorHandler = function errorHandler(
...
+      ...err,
       message: 'Validation failed',
-      details: err?.fields,
-    }
-    return res.status(422).json(response)
+    })
   }
   if (err instanceof HttpResponse) {
-    logger.debug(`Bad request for ${req.path}`)
+    logger.error('Unexpected error thrown in handler: %s', err.message)
```
jonmattgray, line 72: some of the errors are expected? Like a `400`

src/lib/services/identity.ts
```diff
@@ -12,7 +12,7 @@ const getMemberByAlias = async (alias: string) => {
   }
 
   if (res.status === 404) {
-    throw new NotFound(`Member "${alias}" does not exist`)
+    throw new NotFound('capacity')
```
jonmattgray, line 15: it's the identity that hasn't been found, not the capacity

src/models/attachment.ts
```diff
@@ -0,0 +1,24 @@
...
+ *     "type": "string"
+ *   },
+ *   "datetime": "2023-03-16T13:18:42.357Z"
+ * }]
+ */
+export interface Attachment {
+  /**
+   * uuid generated using knex
+   */
+  id?: string
```
jonmattgray, line 17: use `UUID` type?

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      return attachment
+    }
+
+    // no need to check for filename since JSON.parse will throw
+    if (type === 'json' || accept === 'application/json') {
+      this.setHeader('content-type', 'application/json')
+      return JSON.parse(attachment.binary_blob)
+    }
+
+    throw new BadRequest()
```
jonmattgray, line 92: should be a `500` if it reaches here?

jonmattgray: not sure, is it always the case that if it reaches this point it must be because of a bad request?

src/controllers/attachment/index.ts
n3op2: Thanks for flagging this up. it has been updated.

src/controllers/attachment/index.ts
n3op2: 500 along with internal server error message (does not include stack trace)

So we check for http error first, then for exception which is below.

```ts
  if (err instanceof Error) {
    logger.error('Unexpected error thrown in handler: %s', err.message)

    return res.status(500).json(err)
  }
```

src/controllers/attachment/index.ts
n3op2: updated. need to improve this as it looks like causing a lot of back and forth

src/controllers/attachment/index.ts
n3op2: yes! this is why needed to install multer as dev dep. However it did not like both type e.g. multipart/json

src/lib/error-handler/index.ts
n3op2: not sure if I understand this question. In this instance we do not want to log debug because this is an error even tho  'client-side' error. why we are logging as erorr? when it comes to millions of logs entries we want to be able to find all errors and trace our request back if needed. This will allow us to 'filter error.level = 40/50/30/10/etc'

The req_id and levels do help as we can filter log by level e.g. error = 50 so we can say if there was an error.level = 50 over 24hours - we do something e.g. send a slack notification so it can be investigated, or we put to messaging service marked as 'un-processes' whatever:)

src/lib/services/identity.ts
n3op2: it has been updated, thanks for a flag up, find and replace not always the best friend

src/models/attachment.ts
n3op2: created before you created and it has been updated.

src/controllers/attachment/index.ts
n3op2: due to it being a GET request I think 400 is fair to return knowing that parameter was off. e.g. we did not supply a valid ID and etc. It will return 500 when it hits a server error .e.g json.parse fails, db table does not exists. 

BadRequest - or any 4XX indicates normally a client error, body missing a property, request not good enough etc.

src/lib/error-handler/index.ts
jonmattgray: I mean that any HttpResponse error is an error we've defined so it's not unexpected like a `500`, it's an expected error, but it's only a logger message so very minor point

src/lib/services/identity.ts
```diff
@@ -12,7 +12,7 @@ const getMemberByAlias = async (alias: string) => {
   }
 
   if (res.status === 404) {
-    throw new NotFound(`Member "${alias}" does not exist`)
+    throw new NotFound('identity')
```
jonmattgray, line 15: is this not making the error message less helpful, before it would you tell you which alias wasn't found

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!req.body) throw new Error('nothing to upload') // TODO return correct (badreq)
+
+    await this.db.attachment().insert({
+      filename: 'json',
+      binary_blob: Buffer.from(JSON.stringify(req.body)),
+    })
+
+    return 'success'
```
jonmattgray, line 57: don't we want it to return the ID so we can use the attachment in some other request?

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+        binary_blob: Buffer.from(file.buffer),
+      })
+    }
+
+    if (!req.body) throw new Error('nothing to upload') // TODO return correct (badreq)
+
+    await this.db.attachment().insert({
+      filename: 'json',
+      binary_blob: Buffer.from(JSON.stringify(req.body)),
+    })
```
jonmattgray, line 55: when `file` is true it inserts the attachment twice. the `json` insert always happens

src/controllers/attachment/index.ts
jonmattgray: how do you upload json on swagger? Only option is multipart
image

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import { logger } from '../../lib/logger'
+import Database, { Models, Query } from '../../lib/db'
+import type { Attachment } from '../../models'
+import { BadRequest, NotFound } from '../../lib/error-handler'
+
+type File = {
+  [k: string]: string
+}
+
+@Route('attachment')
```
jonmattgray, line 14: still missing the @Tags to give its own section in swagger

src/controllers/attachment/index.ts
mattdean-digicatapult: Yeah this is a 400 not a 500. Bad user request not our fault

src/lib/services/identity.ts
mattdean-digicatapult: Hang on this is actually worse than this. `alias` isn't a path parameter so this should be a `400` not a `404`

src/controllers/attachment/index.ts
n3op2: Yea, it was defaulting to 'default' added a @Tags decorator.
Screenshot 2023-03-17 at 13 07 36

src/controllers/attachment/index.ts
n3op2: Looks like forgot a return statement after performing a db method
```ts
    if (file) {
      await this.db
        .attachment()
        .insert({
          filename: file.originalname,
          binary_blob: Buffer.from(file.buffer),
        })
        .then((row: Array<unknown>) => row[0]) // return id implement along with db changes since it already addresses this

      return 'attachment has been created'
    }
```

src/controllers/attachment/index.ts
n3op2: This has been updated

src/lib/services/identity.ts
n3op2: it will tell you, we just need to type less characters 
Instead of writting all message we can 'peak into' the TS interface **NotFound** and it will return the message, is just that we need to type less as it will -> 

Happy to update not found to `does not exist`
```ts

  constructor(item: Entities) {
    super({ code: 404, message: `${item} not found` })
    this.item = item
    this.name = 'not found'
    // this.stack = (<any> new Error()).stack
  }
```

mattdean-digicatapult: still is not a 404

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,117 @@
...
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachment has been created')
+  public async create(@Request() req: express.Request, @UploadedFile() file: File): Promise<string> {
+    this.log.debug(`creating an attachment ${JSON.stringify(file || req.body)}`)
+    if (file) {
+      const id = await this.db
```
jonmattgray, line 59: `id` is undefined when I just tried creating an attachment

src/lib/services/identity.ts
jonmattgray: I'm just trying to make our lives easier when demoing, identity setting can be confusing and easy to forget

src/controllers/attachment/index.ts
n3op2: updating as we speak, thanks for a check up

n3op2: needed to add to row[0].<prop_name> 
This has been updated and test results can be seen below:
Screenshot 2023-03-17 at 14 09 52

src/lib/services/identity.ts
n3op2: point taken and address should be visible with the latest commit. Changed it to be a string.

Once we have all entities known , then we can update this to be strongly typed e.g. not just any string, but one of... 'table', 'chair', 'wall'...

At the moment it can be any string, ideally we would like to translate models into available items if it makes sense. The following PR

n3op2: This has been updated along with Entities so it's no longer an Entity - instead it can be any string - (for time being) already worried about the merge.

src/controllers/attachment/index.ts
mattdean-digicatapult: what if someone uploads a file with a name of `json`? That is a 400 error. You should have asked for an `application/octet-stream`

src/lib/db/index.ts
mattdean-digicatapult: Is this happening for this PR or not?

src/controllers/attachment/index.ts
mattdean-digicatapult: this is very annoying and is a regression from what we had before. Sadly it appears you're right @n3op2 and tsoa isn't capable of this based off of https://github.com/lukeautry/tsoa/issues/968. I'm honestly not sure what's best as open api does allow this behaviour and the tsoa guys have just decided it's not ok without giving reasons

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const id = await this.db
+      .attachment()
+      .insert({
+        filename: 'json',
+        binary_blob: Buffer.from(JSON.stringify(req.body)),
+      })
+      .then((row: Array<unknown>) => row[0]) // return id implement along with db changes since it already addresses this
+
+    return `${id} JSON attachment has been created`
```
mattdean-digicatapult, line 81: also why are we just returning a string?

mattdean-digicatapult: presviously we returned an AttachmentEntry https://github.com/digicatapult/inteli-api/blob/7fe3fbc2edc68d3483fcfde251f6fd6b610048c3/app/api-v1/routes/attachment.js#L69

src/controllers/attachment/index.ts
n3op2: this is ok, because we check for headers, so I would assume that client will pass the correct header depending on a file. e.g. it's JSON so we will say it's JSON by simply being able to JSON.parse(req.body). If it's a file we will have req.file property.

In this case:
- we check if req.file exists (due to express and multer it will be appended to the req.file)
- we check if req.body exists since we checked if req.file exists and since it does not we attempt JSON (due to body parser JSON will be mapped to req.body)
- if JSON.parse fails- we return 500 along with the err obj.

n3op2: Mmm... in regards to bad request error this is something I just came up with, happy to update with any other or/and add new error object for invalid media type.

src/controllers/attachment/index.ts
n3op2: Mmm... any ideas how to move forward? As a work around I can do something like I did for `getById()` route. We can pass a header via curl and it will work. But that means that there will be two possible routes e.g.

```ts
  public async getById(
    ...
    @Header('return-type') type: 'json' | 'file'
    ...
    if (type === 'file' || accept === 'application/octet-stream') {
    return ...
```

package.json
n3op2: we are not, but TSOA is using for `@UploadedFile()` descriptor. It utilizes in a router.ts

P.S the below is generated by TSOA
```ts
// @ts-ignore - no great way to install types from subpackage
const promiseAny = require('promise.any');
import type { RequestHandler } from 'express';
import * as express from 'express';
const multer = require('multer');
const upload = multer();

        app.post('/attachment',
            authenticateMiddleware([{"bearerAuth":[]}]),
            upload.single('file'),
            ...(fetchMiddlewares<RequestHandler>(attachment)),
            ...(fetchMiddlewares<RequestHandler>(attachment.prototype.create)),
...
```

src/lib/db/index.ts
n3op2: No

package.json
mattdean-digicatapult: ok so this should be a dpendency not a devdependency

src/controllers/attachment/index.ts
mattdean-digicatapult: I'm not sure you get my point. A file could be called `json` that is legitimately not json. In that case ideally we'd just return an octet stream and not throw at all

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,120 @@
...
+    return {
+      attachment: await this.db
+        .attachment()
+        .insert({
+          filename: 'json',
+          binary_blob: Buffer.from(JSON.stringify(req.body)),
+        })
+        .then((row: Array<unknown>) => row[0]),
+      message: 'JSON attachment has been created',
+    }
```
mattdean-digicatapult, line 83: Again this doesn't match the attachment type from the list api

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (!req.body) throw new BadRequest('nothing to upload')
+
+    return {
+      attachment: await this.db
+        .attachment()
+        .insert({
+          filename: 'json',
+          binary_blob: Buffer.from(JSON.stringify(req.body)),
+        })
+        .then((row: Array<unknown>) => row[0]),
```
mattdean-digicatapult, line 81: also don't return the blob. This should be obvious from the `Attachment` type

package.json
n3op2: this should have been updated now

src/controllers/attachment/index.ts
n3op2: this should also be address with the a tmp 'select', but as we talked earlier will be updated to generic get/getBy/create wrappers manufactured for each db entity

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    return result.map(({ binary_blob, ...item }: Attachment) => ({
+      ...item,
+      size: binary_blob.size,
+    }))
+  }
+
+  @Post('/')
+  @SuccessResponse(201, 'attachment has been created')
+  public async create(@Request() req: express.Request, @UploadedFile() file: File): Promise<{ [k: string]: string }> {
```
mattdean-digicatapult, line 56: this should be able to just return an `Attachment` right?

src/controllers/attachment/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  ): Promise<Attachment> {
+    this.log.debug(`attempting to retrieve ${id} attachment`)
+    const { accept } = req.headers
+    const [attachment] = await this.db.attachment().where({ id })
+    if (!attachment) throw new NotFound('attachment')
+    const { filename, binary_blob } = attachment
+
+    if (type === 'json' || accept === 'application/json') {
+      this.setHeader('content-type', 'application/json')
+      return JSON.parse(binary_blob)
```
mattdean-digicatapult, line 99: this behaviour is still incorrect. If `JSON.parse(binary_blob)` throws we should catch and return it as an `application/octet-stream`

src/controllers/attachment/index.ts
n3op2: Mmm.. yes according to knex [insert](https://knexjs.org/guide/query-builder.html#insert) and [returning](https://knexjs.org/guide/query-builder.html#returning) docs added returning most of the props, probly could be taken from payload, but i think from db makes more sense.

This is another reason why we need to have generic wrappers for ORM e.g. getBy etc

src/controllers/attachment/index.ts
n3op2: this has been addressed

src/controllers/attachment/index.ts
```diff
@@ -87,25 +95,23 @@ export class attachment extends Controller {
     @Request() req: express.Request,
     @Path() id: string,
     @Header('return-type') type: 'json' | 'file'
-  ): Promise<Attachment> {
+  ): Promise<Attachment | Blob> {
```
mattdean-digicatapult, line 98: This doesn't return an `Attachment | Blob` it returns `any | Blob`

n3op2: Thanks for a flag up, indeed we are returning whatever JSON is! Updated to be an object since we use JSON.parse() it has to be an `JSON`, right? Happy to change to `<any>`

PR #17, opened 2023-03-16
POST `/capacity/{capacityId}/creation`

jonmattgray:


.depcheckrc
```diff
@@ -1,4 +1,5 @@
 ignores: [
+  "@digicatapult/dscp-process-management",
```
dblane-digicatapult, line 2: You can remove this.

package.json
```diff
@@ -33,6 +34,7 @@
   },
   "homepage": "https://github.com/digicatapult/dscp-matchmaker-api#readme",
   "devDependencies": {
+    "@digicatapult/dscp-process-management": "^1.6.10",
```
dblane-digicatapult, line 37: Again, we don't need this as an installed dependency.  `npx's` default behaviour is to run a command if its installed in the local package directory and then to download and run it if it doesn't exist.

dblane-digicatapult: Or is it required for tests?

test/integration/capacity.test.ts
```diff
@@ -58,12 +63,34 @@ describe('capacity', () => {
...
-        id: capacityId,
+        id: seededCapacityId,
         owner: selfAlias,
-        status: DemandStatus.Created,
+        state: DemandState.created,
         parametersAttachmentId,
       })
     })
+
+    test('it should create a capacity on-chain', async () => {
```
mattdean-digicatapult, line 73: I think we need a bit more test coverage. What happens if the `runProcess` call fails?

mattdean-digicatapult (commented):
I'm not entirely convinced we'll be able to reconcile state with the current `dscp-api` but that's not this ticket. Need a bit more test coverage I think but we're looking pretty close I think

package.json
jonmattgray: I'll remove

.depcheckrc
jonmattgray: removed

test/integration/capacity.test.ts
jonmattgray: is this enough, `500` when `run-process` throws an error https://github.com/digicatapult/dscp-matchmaker-api/pull/17/commits/5ea1e024b366e1a286a3243e13427e81921fe576?

mattdean-digicatapult (approved):

PR #21, opened 2023-03-20
Accept workflow

mattdean-digicatapult:
* Combined Accept workflow
* Combined AcceptFinal workflow

### DIFF
--- a/processFlows.json
+++ b/processFlows.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "Demand - Create",
+    "name": "demand-create",
     "version": 1,
     "program": [
       {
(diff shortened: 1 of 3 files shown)


processFlows.json
```diff
@@ -558,6 +549,747 @@
       },
       { "op": "And" },
       { "op": "Or" },
+      { "op": "And" },
+      {
+        "restriction": {
+          "OutputHasMetadata": {
+            "index": 0,
```
jonmattgray, line 556: `index: 2`? If this is for the new `MATCH2` token?

mattdean-digicatapult: if you look this is adding `originalId` prevention to match2-create. Should have been there before really

mattdean-digicatapult: yep you're right

PR #23, opened 2023-03-21
Match2 table + model

jonmattgray:
Add the migration for a `match2` table and its types

image

### DIFF
--- a/processFlows.json
+++ b/processFlows.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "demand-create",
+    "name": "demand-propose",
     "version": 1,
     "program": [
       {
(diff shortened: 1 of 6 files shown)


processFlows.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "name": "demand-create",
+    "name": "demand-propose",
```
mattdean-digicatapult, line 3: ?

jonmattgray: the token has `state: proposed`, was going for consistency. Could go other way and make `state: created`

mattdean-digicatapult: I think we should have state created, but let me do it in my upcoming PR

mattdean-digicatapult (approved):

PR #27, opened 2023-03-21
Bug/fix json attachment

mattdean-digicatapult:
This is ugly but I think it works. Correctly changes the open-api spec to support both `application/json` and `miltipart/form-data` on upload and then uses the accept header to determine download

### DIFF
--- a/src/controllers/attachment/index.ts
+++ b/src/controllers/attachment/index.ts
@@ -8,9 +8,9 @@ import {
   Route,
   Path,
   Post,
-  Header,
   Response,
   SuccessResponse,
+  Produces,
 } from 'tsoa'
 import { Logger } from 'pino'
 import express from 'express'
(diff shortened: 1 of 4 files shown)


src/controllers/attachment/index.ts
```diff
@@ -91,28 +124,27 @@ export class attachment extends Controller {
...
-    // log and default to octect-stream
-    if (type === 'json' || accept === 'application/json') {
-      this.setHeader('content-type', 'application/json')
-      try {
-        return JSON.parse(binary_blob)
-      } catch (err) {
-        this.log.warn('requested type failed, returning as octet')
+    const orderedAccept = parseAccept(req.headers.accept || '*/*')
+    if (filename === 'json') {
+      for (const mimeType of orderedAccept) {
```
n3op2, line 138: why do we need this logic? we are kinda already documented headers. I think we are doubling up here on checks or I'm missing something

src/controllers/attachment/index.ts
```diff
@@ -91,28 +124,27 @@ export class attachment extends Controller {
...
     const { filename, binary_blob } = attachment
 
-    // log and default to octect-stream
-    if (type === 'json' || accept === 'application/json') {
-      this.setHeader('content-type', 'application/json')
-      try {
-        return JSON.parse(binary_blob)
-      } catch (err) {
-        this.log.warn('requested type failed, returning as octet')
+    const orderedAccept = parseAccept(req.headers.accept || '*/*')
```
n3op2, line 136: under what instance we would get accept headers containing multiple. Is this is required by other parts of DSCP?

src/controllers/attachment/index.ts
```diff
@@ -91,28 +124,27 @@ export class attachment extends Controller {
...
+    if (filename === 'json') {
+      for (const mimeType of orderedAccept) {
+        if (mimeType === 'application/json' || mimeType === 'application/*' || mimeType === '*/*') {
+          const json = JSON.parse(binary_blob)
+          return json
+        }
+        if (mimeType === 'application/octet-stream') {
```
n3op2, line 143: like this check looks useless since we are going to return `return this.octetResponse(binary_blob, filename)` response

src/controllers/attachment/index.ts
mattdean-digicatapult: it's part of the Accept header standard https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

src/controllers/attachment/index.ts
mattdean-digicatapult: we're not doubling up at all. This is the same logic as from `inteli-api` as per the ticket requirements. This logic replaces the logic on `return-type` which is a duplicate headerin meaning of `Accept`. I may have missed some exception handling here but I think this is closer

src/controllers/attachment/index.ts
mattdean-digicatapult: We're in a loop so no it's not useless. If some one asks for a random mimetype (not covered here) and as a second preference an `application/json` response you would not want it to return an octet stream

n3op2 (dismissed):

PR #28, opened 2023-03-22
Tests/l3 25

n3op2:
# What

A follow up PR including integration test coverage for attachment controller. 

### Scenarios:

```sh
  attachment
    ✔ attachment as octect with the filename [json] (106ms)
    invalid requests
      ✔ returns 500 when attempting to retrieve not UUID
      ✔ returns 404 if no records found
    uploads and retrieves attachment
      ✔ confirms JSON and octet attachment uploads
      ✔ returns octet attachment
      ✔ returns JSON attachment
      ✔ returns octet when JSON.parse fails


  7 passing (530ms)
```

### DIFF
--- a/test/integration/attachment.test.ts
+++ b/test/integration/attachment.test.ts
@@ -73,11 +73,19 @@ describe('attachment', () => {
 
     test('returns octet attachment', async () => {
       const { id } = octetRes.body
-      const { status, body } = await get(app, `/attachment/${id}`, { accept: 'application/octet-stream' })
+      const { status, body, header } = await get(app, `/attachment/${id}`, { accept: 'application/octet-stream' })
 
       expect(status).to.equal(200)
       expect(body.type).to.equal('Buffer')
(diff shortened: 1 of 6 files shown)


test/integration/attachment.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  before(async () => {
+    app = await createHttpServer()
+  })
+
+  afterEach(async () => {
+    await db.attachment().del()
+  })
+
+  describe('invalid requests', () => {
+    test('returns 500 when attempting to retrieve not UUID', async () => {
```
jonmattgray, line 28: if the route controller changes `@Path() id: string,` -> `@Path() id: UUID` then OpenAPI will catch that it's not a UUID and return a 400. And doesn't let you put a non-UUID on Swagger
image

mattdean-digicatapult (approved):

jonmattgray (approved):

PR #31, opened 2023-03-22
`POST /match2` + `GET /match2` + `GET /match2/{id}`

jonmattgray:
Add the following routes with tests:
image

### DIFF
--- a/src/controllers/match2/index.ts
+++ b/src/controllers/match2/index.ts
@@ -34,11 +34,12 @@ export class Match2Controller extends Controller {
   }
 
   /**
-   * A Member creates a new Match2 for an order and a capacity by referencing each demand.
-   * @summary Create a new Match2
+   * A Member creates a new match2 for an order and a capacity by referencing each demand.
+   * @summary Create a new match2
    */
   @Post()
   @Response<BadRequest>(400, 'Request was invalid')
+  @Response<ValidateError>(422, 'Validation Failed')
   @SuccessResponse('201')
   public async createMatch2(
     @Body() { demandA: demandAId, demandB: demandBId }: Match2Request
@@ -73,7 +74,7 @@ export class Match2Controller extends Controller {
 
   /**
    * Returns the details of all match2s.
-   * @summary List all capacity demands
+   * @summary List all match2s
    */
   @Get('/')
   public async getAll(): Promise<Match2Response[]> {
--- a/src/controllers/match2/index.ts
+++ b/src/controllers/match2/index.ts
@@ -1,0 +1,116 @@
+import {
+  ValidateError,
+  Controller,
+  Post,
+  Get,
+  Route,
+  Response,
+  Body,
+  SuccessResponse,
+  Tags,
+  Security,
+  Path,
+} from 'tsoa'
+import type { Logger } from 'pino'
+
+import { logger } from '../../lib/logger'
+import Database from '../../lib/db'
+import { BadRequest, NotFound } from '../../lib/error-handler/index'
+import { getMemberByAddress, getMemberBySelf } from '../../lib/services/identity'
+import { Match2Request, Match2Response, Match2State } from '../../models/match2'
+import { UUID } from '../../models/uuid'
+
+@Route('match2')
+@Tags('match2')
+@Security('bearerAuth')
+export class Match2Controller extends Controller {
+  log: Logger
+  db: Database
+
+  constructor() {
+    super()
+    this.log = logger.child({ controller: '/match2' })
+    this.db = new Database()
+  }
+
+  /**
+   * A Member creates a new Match2 for an order and a capacity by referencing each demand.
+   * @summary Create a new Match2
+   */
+  @Post()
+  @Response<BadRequest>(400, 'Request was invalid')
+  @SuccessResponse('201')
+  public async createMatch2(
+    @Body() { demandA: demandAId, demandB: demandBId }: Match2Request
+  ): Promise<Match2Response> {
+    const [demandA] = await this.db.getDemand(demandAId)
+    if (!demandA) {
+      throw new BadRequest('Demand A not found')
+    }
+
+    const [demandB] = await this.db.getDemand(demandBId)
+    if (!demandB) {
+      throw new BadRequest('Demand B not found')
+    }
+
+    if (demandA.subtype === demandB.subtype) {
+      throw new BadRequest(`Demands have matching type: ${demandA.subtype}`)
+    }
+
+    const selfAddress = await getMemberBySelf()
+
+    const [match2] = await this.db.insertMatch2({
+      optimiser: selfAddress,
+      member_a: demandA.owner,
+      member_b: demandB.owner,
+      state: Match2State.proposed,
+      demand_a_id: demandAId,
+      demand_b_id: demandBId,
+    })
+
+    return responseWithAliases(match2)
+  }
+
+  /**
+   * Returns the details of all match2s.
+   * @summary List all capacity demands
+   */
+  @Get('/')
+  public async getAll(): Promise<Match2Response[]> {
+    const match2s = await this.db.getMatch2s()
+    const result = await Promise.all(match2s.map(async (capacity: Match2Response) => responseWithAliases(capacity)))
+    return result
+  }
+
+  /**
+   * @summary Get a match2 by ID
+   * @param match2Id The match2's identifier
+   */
+  @Response<ValidateError>(400, 'Validation Failed')
+  @Response<NotFound>(404, 'Item not found')
+  @Get('{match2Id}')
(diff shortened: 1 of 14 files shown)


src/controllers/match2/index.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  @Post()
+  @Response<BadRequest>(400, 'Request was invalid')
+  @Response<ValidateError>(422, 'Validation Failed')
+  @SuccessResponse('201')
+  public async proposeMatch2(
+    @Body() { demandA: demandAId, demandB: demandBId }: Match2Request
+  ): Promise<Match2Response> {
+    const [demandA] = await this.db.getDemand(demandAId)
+    if (!demandA) {
```
n3op2, line 48: just wonder if it we could look for both ids e.g. demandAId and demandBId with something like SQL based in() where instead of haing to duplicate if statements.

It doe not stop me from approving. I just believe that less is more:)

src/controllers/match2/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * @summary Get a match2 by ID
+   * @param match2Id The match2's identifier
+   */
+  @Response<ValidateError>(400, 'Validation Failed')
+  @Response<NotFound>(404, 'Item not found')
+  @Get('{match2Id}')
+  public async getCapacity(@Path() match2Id: UUID): Promise<Match2Response> {
+    const [match2] = await this.db.getMatch2(match2Id)
+    if (!match2) throw new NotFound('Match2 Not Found')
```
n3op2, line 95: since it's not found can we just passed the name of a model and prefix 'not found?' again this is something I think makes easier to understand a code for a person who just recently started

src/controllers/match2/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    return responseWithAliases(match2)
+  }
+}
+
+const responseWithAliases = async (match2: Match2Response): Promise<Match2Response> => {
+  const [{ alias: optimiser }, { alias: memberA }, { alias: memberB }] = await Promise.all([
+    getMemberByAddress(match2.optimiser),
```
n3op2, line 103: this is something where I would look to abstract or update a controller on identity service. Again just aesthetics

src/lib/db/index.ts
```diff
@@ -14,6 +14,28 @@ export interface Models<V> {
 
 export type Query = Knex.QueryBuilder
 
+const demandColumns = [
+  'id',
+  'owner',
+  'state',
+  'subtype',
+  'parameters_attachment_id AS parametersAttachmentId',
+  'latest_token_id AS latestTokenId',
```
n3op2, line 23: this is where I hope that we could get to the point where we have models like in prisma -> https://www.prisma.io/docs/concepts/components/prisma-schema/data-model, However, wanted to find some knex references but was not able to validate, so far got this -> https://knexjs.org/guide/#postprocessresponse. [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller)

src/lib/db/index.ts
```diff
@@ -14,6 +14,28 @@ export interface Models<V> {
...
+  'original_token_id AS originalTokenId',
+]
+
+const match2Columns = [
```
n3op2, line 27: same as above comment, I guess it's a matter of how quick or how scalable

src/lib/db/index.ts
```diff
@@ -84,4 +103,16 @@ export default class Database {
       })
       .where({ id: localId })
   }
+
+  insertMatch2 = async (match2: object) => {
```
n3op2, line 107: love how it exponentially grows. This is why I would suggest to start utilising models and MVC approach within a TS env. Otherwise we are just duplicating intelli-api

src/controllers/match2/index.ts
n3op2: we perform 3 request and not even in parellel

src/controllers/match2/index.ts
```diff
@@ -0,0 +1,117 @@
...
+@Tags('match2')
+@Security('bearerAuth')
+export class Match2Controller extends Controller {
+  log: Logger
+  db: Database
+
+  constructor() {
+    super()
+    this.log = logger.child({ controller: '/match2' })
+    this.db = new Database()
```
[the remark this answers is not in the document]
jonmattgray: yeah, I guess that ticket you wrote will address this whenever we do it

src/controllers/match2/index.ts
jonmattgray: I will add. It will mean another `db` method though!

jonmattgray: which you may not like

src/controllers/match2/index.ts
jonmattgray: changed to just `('match2')`

src/controllers/match2/index.ts
jonmattgray: after trying it out it kind of ends up being more complicated getting them both at the same time, cause you end up checking length and figuring out which one was wrong

src/controllers/match2/index.ts
jonmattgray: what kind of abstraction are you thinking?

src/lib/db/index.ts
jonmattgray: yes the case swapping is not ideal

test/integration/match2.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      expect(response.body).to.equal('Demands have matching type: order')
+    })
+
+    test('both demands are capacities - 400', async () => {
+      const response = await post(app, '/match2', { demandA: seededCapacityId, demandB: seededCapacityId })
+      expect(response.status).to.equal(400)
+      expect(response.body).to.equal('Demands have matching type: capacity')
+    })
+
+    test('invalid demand uuid - 422', async () => {
```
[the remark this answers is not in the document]
jonmattgray: we seem to use both across our repos. I've changed everything here to `it` because it was fewer changes to `main`

test/integration/match2.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      const response = await post(app, '/match2', { demandA: 'invalid', demandB: seededCapacityId })
+      expect(response.status).to.equal(422)
+      expect(response.body.message).to.equal('Validation failed')
+    })
+
+    it('non-existent match2 id - 404', async () => {
```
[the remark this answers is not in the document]
jonmattgray: changed to all use `it`

src/lib/db/index.ts
jonmattgray: https://digicatapult.atlassian.net/browse/L3-52 will sort it?

mattdean-digicatapult: There are better ways of doing this granted. It doesn't grow expontentially though it grows linearly. Sorry minor bugbear there

mattdean-digicatapult (changes requested):
1 change that must be made I think. The rest looks ok

src/controllers/match2/index.ts
jonmattgray: no it was always `Promise.all`

src/controllers/match2/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (!demandA) {
+      throw new BadRequest('Demand A not found')
+    }
+
+    const [demandB] = await this.db.getDemand(demandBId)
+    if (!demandB) {
+      throw new BadRequest('Demand B not found')
+    }
+
+    if (demandA.subtype === demandB.subtype) {
```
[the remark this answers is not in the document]
jonmattgray: thanks, missed that on the process flow, I thought it could be either way round. Added a stricter check and updated tests https://github.com/digicatapult/dscp-matchmaker-api/pull/31/commits/61b4c7104eff4cf017108351e771d2ee88bb9da3

mattdean-digicatapult (approved):
lgtm

src/lib/db/index.ts
n3op2: @jonmattgray on that note we are already looking into an ORM and I think this might become a little irrelevant for time being. However, the logic would remain the same. We do lie ORM does, create wrappers for psql bu in our case is a framework we use https://github.com/digicatapult/dscp-matchmaker-api/blob/40774f372de43f1d6f598ad98c95abd0652a673b/src/lib/db/index.ts <- this is an example or the idea of 'factory'

test/integration/match2.test.ts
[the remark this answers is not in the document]
n3op2: nice, i did as well!

n3op2 (approved):

PR #32, opened 2023-03-23
L3-71 Get capacity creation

nitro-marky:
Gets Creation ID and includes a test.


src/controllers/.DS_Store
dblane-digicatapult, line 1: Delete

test/integration/capacity.test.ts
```diff
@@ -124,5 +124,10 @@ describe('capacity', () => {
       const response = await post(app, `/capacity/${seededCapacityId}/creation`, {})
       expect(response.status).to.equal(500)
     })
+
+    it('non-existant Creation ID', async () => {
```
jonmattgray, line 128: need a happy path test too

src/lib/db/index.ts
```diff
@@ -84,4 +84,8 @@ export default class Database {
       })
       .where({ id: localId })
   }
+
+  getCreationID = async (capacityId: UUID) => {
```
jonmattgray, line 88: rather than getting a creation ID, this is getting a creation transaction? Also the ID argument is a `creationId` or `transactionId` rather than ` capacityId`

PR #33, opened 2023-03-23
Ls 71

nitro-marky:
GET capacity/{capacity_id}/creation/{creation_id} with happy and sad test.

### DIFF
--- a/src/lib/db/index.ts
+++ b/src/lib/db/index.ts
@@ -82,7 +82,7 @@ export default class Database {
   }
 
   getTransaction = async (id: UUID) => {
-    return this.db().transaction().select('*').where({ id })
+    return this.db().transaction().select('*').where({ id: id })
   }
 
   updateTransaction = async (transactionId: UUID, transaction: object) => {
--- a/src/lib/db/index.ts
+++ b/src/lib/db/index.ts
@@ -85,10 +85,6 @@ export default class Database {
     return this.db().transaction().select('*').where({ id })
   }
 
-  getTransactions = async () => {
-    return this.db().transaction().select('*')
-  }
-
   updateTransaction = async (transactionId: UUID, transaction: object) => {
     return this.db()
       .transaction()
--- a/src/lib/db/index.ts
+++ b/src/lib/db/index.ts
@@ -66,6 +66,10 @@ export default class Database {
     return this.db().transaction().select('*').where({ id })
   }
 
+  getTransactions = async () => {
+    return this.db().transaction().select('*')
+  }
+
   updateTransaction = async (transactionId: UUID, transaction: object) => {
     return this.db()
       .transaction()
--- a/test/integration/capacity.test.ts
+++ b/test/integration/capacity.test.ts
@@ -142,8 +142,13 @@ describe('capacity', () => {
       expect(response.status).to.equal(500)
     })
 
-    it('non-existant Creation ID', async () => {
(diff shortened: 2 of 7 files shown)


test/integration/capacity.test.ts
```diff
@@ -90,8 +90,17 @@ describe('capacity', () => {
...
+      identitySelfMock()
+      const response = await get(app, `/capacity/${seededCapacityId}/creation/${seededTransactionId}`)
+      expect(response.status).to.equal(200)
+    })
+
   })
 
+
+
   describe('sad path', () => {
```
n3op2, line 104: 'sad path' - :))))

n3op2 (approved):

src/lib/db/index.ts
```diff
@@ -85,6 +85,10 @@ export default class Database {
     return this.db().transaction().select('*').where({ id })
   }
 
+  getTransactions = async () => {
+    return this.db().transaction().select('*')
+  }
+
```
jonmattgray, line 91: function isn't used anywhere?

test/integration/capacity.test.ts
```diff
@@ -90,8 +90,17 @@ describe('capacity', () => {
       expect(capacity.latestTokenId).to.equal(mockTokenId)
       expect(capacity.originalTokenId).to.equal(mockTokenId)
     })
+
+    it('it should get transactions', async () => {
+      identitySelfMock()
```
jonmattgray, line 95: shouldn't be required as requesting a transaction doesn't hit the identity service

test/integration/capacity.test.ts
```diff
@@ -120,5 +129,10 @@ describe('capacity', () => {
       const response = await post(app, `/capacity/${seededCapacityId}/creation`, {})
       expect(response.status).to.equal(500)
     })
+
+    it('non-existant Creation ID', async () => {
```
jonmattgray, line 133: there could be another sad path test for the other 404 error the route throws

test/integration/capacity.test.ts
```diff
@@ -90,8 +90,17 @@ describe('capacity', () => {
       expect(capacity.latestTokenId).to.equal(mockTokenId)
       expect(capacity.originalTokenId).to.equal(mockTokenId)
     })
+
+    it('it should get transactions', async () => {
+      identitySelfMock()
+      const response = await get(app, `/capacity/${seededCapacityId}/creation/${seededTransactionId}`)
```
jonmattgray, line 96: good idea to assert the contents of `response.body` too to make sure it's returning what we want

n3op2: yea, I would say even good to assert all properties so then we make a change test will flag up.

test/integration/capacity.test.ts
n3op2: this is for phuture

jonmattgray: what?

jonmattgray: what is phuture?

test/seeds/index.ts
```diff
@@ -39,6 +41,16 @@ export const seed = async () => {
     },
   ])
 
+  await db.transaction().insert([
+    {
+      id: seededTransactionId,
+      token_type: 'DEMAND',
```
jonmattgray, line 47: you can import the `TokenType` enum and use that instead of writing out the string, same with `state`

jonmattgray (approved):

PR #34, opened 2023-03-23
`POST match2/{id}/proposal` + `GET match2/{id}/proposal`+ `GET match2/{id}/proposal/{id}`

jonmattgray:
Add`POST match2/{id}/proposal` + `GET match2/{id}/proposal`+ `GET match2/{id}/proposal/{id}` routes
Change temp 'blockchain watcher' to not save `token_id` on `transactions`

Match2 token on chain
image

DemandA (order) on chain
image

DemandB (capacity) on chain
image


src/lib/payload.ts
```diff
@@ -15,7 +15,44 @@ export const demandCreate = (demand: DemandPayload, transactionId: UUID) => ({
...
+      metadata: {
+        version: { type: 'LITERAL', value: '1' },
+        type: { type: 'LITERAL', value: TokenType.DEMAND },
+        state: { type: 'LITERAL', value: DemandState.created },
+        subtype: { type: 'LITERAL', value: demandB.subtype },
+        originalId: { type: 'TOKEN_ID', value: demandB.originalTokenId },
+      },
+    },
+    {
+      roles: { Optimiser: match2.optimiser, MemberA: match2.memberA, MemberB: match2.memberB },
```
jonmattgray, line 49: I've found myself using `DemandA.owner` and `DemandB.owner` instead of the roles. Are the roles perhaps redundant? Or maybe I should be checking against them

src/controllers/match2/index.ts
```diff
@@ -101,6 +94,76 @@ export class Match2Controller extends Controller {
...
+  public async createMatch2OnChain(@Path() match2Id: UUID): Promise<TransactionResponse> {
+    const [match2] = await this.db.getMatch2(match2Id)
+    if (!match2) throw new NotFound('match2')
+    if (match2.state !== Match2State.proposed) throw new BadRequest(`Match2 must have state: ${Match2State.proposed}`)
+
+    const [demandA] = await this.db.getDemand(match2.demandA)
+    validatePreOnChain(demandA, DemandSubtype.order, 'DemandA')
+
+    const [demandB] = await this.db.getDemand(match2.demandB)
+    validatePreOnChain(demandB, DemandSubtype.capacity, 'DemandB')
```
n3op2, line 115: since you are creating a helper method for each new db query, who not something like

```js
// pseudo code
const [demandA, demandB] = await this.db.getDemandAB(match2.demandA, match2.demandB)
// and validate
validatePreOnChain(await this.db.getDemandAB(match2.demandA, match2.demandB)
// or ...
await this.db.getDemandAB(match2.demandA, match2.demandB).foEach(demand => validatePreOnChain(....)
...
```

src/controllers/match2/index.ts
```diff
@@ -101,6 +94,76 @@ export class Match2Controller extends Controller {
...
+    })
+
+    // temp - until there is a blockchain watcher, need to await runProcess to know token IDs
+    const tokenIds = await runProcess(match2Propose(match2, demandA, demandB))
+    await this.db.updateTransaction(transaction.id, { state: TransactionState.finalised })
+
+    // match2-propose returns 3 token IDs
+    await observeTokenId(TokenType.DEMAND, match2.demandA, tokenIds[0], false) // order
+    await observeTokenId(TokenType.DEMAND, match2.demandB, tokenIds[1], false) // capacity
+    await observeTokenId(TokenType.MATCH2, match2.id, tokenIds[2], true) // match2
```
n3op2, line 130: three times calling the same function?

src/lib/error-handler/index.ts
```diff
@@ -68,12 +68,12 @@ export const errorHandler = function errorHandler(
     })
   }
   if (err instanceof HttpResponse) {
-    logger.warn('Error thrown in handler: %s', err.message)
+    logger.warn('Error thrown in handler: %s', JSON.stringify(err.message))
```
n3op2, line 71: why do we need to stringify here? Could you please give some examples. The message of error instance is a string.
Screenshot 2023-03-29 at 08 06 24

src/lib/services/identity.ts
```diff
@@ -23,7 +23,7 @@ const getMemberBySelf = async () => {
 
   if (res.ok) {
     const member = await res.json()
-    return member.address
+    return member
```
n3op2, line 26: I genuinely believe less is more, but this is also could be the result of eslint. 
`if (res.of) return await res.json()` - feel free to ignore.

src/models/demand.ts
```diff
@@ -33,6 +33,8 @@ export interface DemandResponse {
 export interface DemandPayload extends DemandResponse, Attachment {
   subtype: DemandSubtype
   binary_blob: Blob
+  latestTokenId: number
```
n3op2, line 36: as it looks we have not updated to be UUIDs and still incrementing?

n3op2 (commented):
Leaving a few comments, nothing major apart from enums being used as values (but that does not trigger to ask for changes) and maybe a few places where I would seek for abstraction.

Did not approve just yet due to a couple of questions. I will approve right after answers. Feel free to not make any changes since I did not request for any.

Also in regards to seeding knex has a nice way of handling seeds -> https://knexjs.org/guide/migrations.html#seed-cli

src/controllers/capacity/index.ts
```diff
@@ -105,6 +104,7 @@ export class CapacityController extends Controller {
   public async createCapacityOnChain(@Path() capacityId: UUID): Promise<TransactionResponse> {
     const [capacity] = await this.db.getDemandWithAttachment(capacityId, DemandSubtype.capacity)
     if (!capacity) throw new NotFound('capacity')
+    if (capacity.state !== DemandState.created) throw new BadRequest(`Demand must have state: ${DemandState.created}`)
```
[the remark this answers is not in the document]
jonmattgray: not sure I understand what you're saying, how does using the `DemandState` enum make it harder to update?

src/controllers/capacity/index.ts
```diff
@@ -139,14 +138,18 @@ export class CapacityController extends Controller {
...
+   */
+  @Response<NotFound>(404, 'Item not found.')
   @SuccessResponse('200')
   @Get('{capacityId}/creation/')
   public async getTransactionsFromCapacity(@Path() capacityId: UUID): Promise<TransactionResponse[]> {
     const [capacity] = await this.db.getDemand(capacityId)
     if (!capacity) throw new NotFound('capacity')
 
-    const creations = await this.db.getTransactionsFromCapacityID(capacityId)
+    const creations = await this.db.getTransactionsByLocalId(capacityId)
```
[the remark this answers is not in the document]
jonmattgray: I'll change to `return await...`

src/controllers/match2/index.ts
```diff
@@ -46,24 +51,12 @@ export class Match2Controller extends Controller {
...
-      throw new BadRequest('Demand B not found')
-    }
-
-    if (demandB.subtype !== DemandSubtype.capacity) {
-      throw new BadRequest(`DemandB must be ${DemandSubtype.capacity}`)
-    }
+    validatePreLocal(demandB, DemandSubtype.capacity, 'DemandB')
 
-    const selfAddress = await getMemberBySelf()
+    const { address: selfAddress } = await getMemberBySelf()
```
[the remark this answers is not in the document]
jonmattgray: I reassigned to `selfAddress` just for clarity when the variable is used later

src/controllers/match2/index.ts
jonmattgray: what I've done is more readable to me, even if it repeats a bit. And I don't need to make a new `DemandAB`  function

src/controllers/match2/index.ts
jonmattgray: it's temporary, pretending to be a blockchain watcher/indexer that's looping over tokens as they appear on chain

src/lib/error-handler/index.ts
jonmattgray: I had some error that was logging as `[Object object]`, I can't remember, it might have been the `dscpApi` error:
```
  const result = await res.json()

  if (res.ok) {
    return result
  }

  throw new HttpResponse({ code: 500, message: result }) // pass through dscpApi error
 ```

src/lib/services/identity.ts
jonmattgray: will change

src/models/demand.ts
jonmattgray: yep still incrementing integer on chain

src/controllers/match2/index.ts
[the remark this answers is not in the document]
jonmattgray: the identity service returns `{address, alias}` to describe any member

src/lib/error-handler/index.ts
jonmattgray: I'll change `dscp-api` and revert the error handler changes

src/controllers/capacity/index.ts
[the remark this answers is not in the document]
n3op2: Nah, I'm saying that more occurrences of (in this instance enum) will make harder to update due to the software being coupled to one or other method/way (in this case enums). In short what I'm saying that it's harder to update when we use it while we build more. So let's say there was one occurrence, not we got more. I see this as a tech debt, so please do not do anything about it. We will use enums as values in this project, there is a backlog story to address that - done. Next one will be better:)

src/controllers/match2/index.ts
n3op2: ok

PR #35, opened 2023-03-24
L3-72 GET capacity/{capacity_id}/creation

nitro-marky:
GET capacity/{capacity_id}/creation + happy and sad test

### DIFF
--- a/src/controllers/capacity/index.ts
+++ b/src/controllers/capacity/index.ts
@@ -144,7 +144,7 @@ export class CapacityController extends Controller {
   @Get('{capacityId}/creation/')
   public async getTransactionsFromCapacity(@Path() capacityId: UUID): Promise<TransactionResponse[]> {
     const capacity = await this.db.getTransactionsFromCapacityID(capacityId)
-    if (!capacity) throw new NotFound('capacity')
+    if (capacity.length === 0) throw new NotFound('capacity')
 
     return capacity
   }
--- a/src/controllers/capacity/index.ts
+++ b/src/controllers/capacity/index.ts
@@ -143,10 +143,13 @@ export class CapacityController extends Controller {
   @SuccessResponse('200')
   @Get('{capacityId}/creation/')
   public async getTransactionsFromCapacity(@Path() capacityId: UUID): Promise<TransactionResponse[]> {
-    const capacity = await this.db.getTransactionsFromCapacityID(capacityId)
+    const [capacity] = await this.db.getDemand(capacityId)
     if (!capacity) throw new NotFound('capacity')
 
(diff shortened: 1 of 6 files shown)


src/controllers/capacity/index.ts
```diff
@@ -138,6 +138,16 @@ export class CapacityController extends Controller {
     if (!creation) throw new NotFound('creation')
     return creation
   }
+
+  @Response<NotFound>(404, 'Items not found.')
+  @SuccessResponse('200')
+  @Get('{capacityId}/creation/')
+  public async getTransactionsFromCapacity(@Path() capacityId: UUID): Promise<TransactionResponse[]> {
+    const capacity = await this.db.getTransactionsFromCapacityID(capacityId)
+    if (capacity.length === 0) throw new NotFound('capacity')
```
jonmattgray, line 147: a length of `0` is ok because there might be no creation transactions for the capacity, you would return `[]`.

However it should `404 - NotFound` if the capacity doesn't exist at all

test/integration/capacity.test.ts
```diff
@@ -109,6 +109,33 @@ describe('capacity', () => {
       )
     })
 
+    it('it should get all transactions from a capacity ID - 200', async () => {
+      const response = await get(app, `/capacity/${seededCapacityId}/creation/`)
+      expect(response.status).to.equal(200)
+      expect(response.body).to.deep.equal(
+        [
+          {
+            id: '1f3af974-7d4d-40b4-86a5-94a2241265cb',
```
jonmattgray, line 118: you can use the exported `seededTransactionId`s for these IDs, also a few enums instead of things like `DEMAND`

src/controllers/capacity/index.ts
```diff
@@ -138,6 +138,19 @@ export class CapacityController extends Controller {
...
+
+  @Response<NotFound>(404, 'Items not found.')
+  @SuccessResponse('200')
+  @Get('{capacityId}/creation/')
+  public async getTransactionsFromCapacity(@Path() capacityId: UUID): Promise<TransactionResponse[]> {
+    const [capacity] = await this.db.getDemand(capacityId)
+    if (!capacity) throw new NotFound('capacity')
+
+    const creations = await this.db.getTransactionsFromCapacityID(capacityId)
+    if (!creations) throw new NotFound('capacity')
```
jonmattgray, line 150: can remove checking if `creations` is falsy, there's no chance of `404`ing at this point because you've already checked the capacity ID. I think the only way `creations` would be a false value like `undefined` would be if there's some random unexpected error in which case the error handler will catch it and response with a `500`

PR #36, opened 2023-03-28
Feature/l3 38

n3op2:
# What

- post controller for a new demand (Order)
- integration tests to validate errors and success

```sh
# integration tests
  order
    ✔ should create a order demand (89ms)
    if attachment can not be found
      ✔ returns 404 along with the message
    if invalid attachment uuid
      ✔ returns 422 along with validation error
```

### Screengrabs
Screenshot 2023-03-28 at 11 16 04

### DIFF
--- a/src/controllers/order/index.ts
+++ b/src/controllers/order/index.ts
@@ -10,20 +10,20 @@ import { getMemberByAddress, getMemberBySelf } from '../../lib/services/identity
 const SUBTYPE = 'order'
 const STATE = 'created'
 
-@Route('demand')
-@Tags('demand')
+@Route('order')
+@Tags('order')
 @Security('bearerAuth')
-export class demand extends Controller {
+export class order extends Controller {
   log: Logger
   db: Database
   constructor() {
     super()
-    this.log = logger.child({ controller: '/demand' })
+    this.log = logger.child({ controller: '/order' })
     this.db = new Database()
   }
 
   /**
-   * A Member creates a new demand for a capacity by referencing an uploaded parameters file.
+   * A Member creates a new demand for a order by referencing an uploaded parameters file.
    * @summary Create a new order demand
    * @param parametersAttachmentId The attachment's identifier
   */
@@ -44,9 +44,10 @@ export class demand extends Controller {
       parameters_attachment_id: parametersAttachmentId,
     })
 
+    console.log({ order })
     return {
       id: order.id,
-      owner: await getMemberByAddress(selfAddress),
+      owner: await getMemberByAddress(selfAddress).then(({ alias }: { alias: string}) => alias),
       state: order.state,
       parametersAttachmentId,
     }
--- a/src/controllers/order/index.ts
+++ b/src/controllers/order/index.ts
@@ -1,14 +1,11 @@
 import { Body, Tags, Security, ValidateError, SuccessResponse, Response, Controller, Post, Route } from 'tsoa'
 import { Logger } from 'pino'
 
-import { DemandRequest, DemandResponse } from '../../models'
+import { DemandRequest, DemandResponse, DemandState, DemandSubtype } from '../../models'
 import { logger } from '../../lib/logger'
 import { BadRequest, NotFound } from '../../lib/error-handler'
 import Database from '../../lib/db'
 import { getMemberByAddress, getMemberBySelf } from '../../lib/services/identity'
-
-const SUBTYPE = 'order'
-const STATE = 'created'
 
 @Route('order')
 @Tags('order')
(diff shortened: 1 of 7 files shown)


src/controllers/order/index.ts
```diff
@@ -0,0 +1,54 @@
...
+   * A Member creates a new demand for a order by referencing an uploaded parameters file.
+   * @summary Create a new order demand
+   * @param parametersAttachmentId The attachment's identifier
+   */
+  @Post()
+  @Response<BadRequest>(400, 'Request was invalid')
+  @Response<NotFound>(404, 'Demand was not found')
+  @Response<ValidateError>(422, 'Validation Failed')
+  @SuccessResponse('201')
+  public async createCapacity(@Body() { parametersAttachmentId }: DemandRequest): Promise<DemandResponse> {
```
jonmattgray, line 35: function name

src/controllers/order/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    const [attachment] = await this.db.getAttachment(parametersAttachmentId)
+    if (!attachment) throw new NotFound('attachment')
```
jonmattgray, line 37: seems like a `400` to me rather than a `404` as it's part of the request body not the URL

src/controllers/order/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const selfAddress = await getMemberBySelf()
```
jonmattgray, line 39: I realised from looking at this that we can use `getMemberSelf()` to get `address` and `alias` at the same time, if you change it to `return member` rather than `member.address` then don't need the second call to identity service to get the owner alias in the `return` block of this function

jonmattgray: so this line would become something like `const { address, alias } = await getMemberBySelf()`

jonmattgray: my mistake from when I did `capacity`

src/controllers/order/index.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Logger } from 'pino'
+
+import { DemandRequest, DemandResponse } from '../../models'
+import { logger } from '../../lib/logger'
+import { BadRequest, NotFound } from '../../lib/error-handler'
+import Database from '../../lib/db'
+import { getMemberByAddress, getMemberBySelf } from '../../lib/services/identity'
+
+const SUBTYPE = 'order'
+const STATE = 'created'
```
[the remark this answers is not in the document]
n3op2: No, unless you give me the reason why. I think we have talked about it. Wanted to update so they are types rather than enums. The validation does not trip and still passes.

src/controllers/order/index.ts
n3op2: thanks! indeed, just blindly thought that since the file name is already an entity we can call it `create`, my bad, it has been updated.

src/controllers/order/index.ts
n3op2: but item is not found, there is another assertion in `order-demand.test.ts` file. @mattdean-digicatapult - thoughts?

It's clearly the attachment id not found and there is another assertion for ID not being UUID where we return validation error.

n3op2: Also it does not mean that it's a bad request, it's good, the item is not found so why we return bad request?

src/controllers/order/index.ts
n3op2: I rather not do it, due to required updates in other places, I think this could be created as another story same as for db module.

src/controllers/order/index.ts
n3op2: Also we assert for request body not being UUID in the integration testing.

jonmattgray: maybe `422` then? `404` suggests the URL path is wrong, which it isn't (on the original request)

n3op2: WHen the URL path is wrong e.g. is not UUID there is a test where it will return ValidateError with 422 https://github.com/digicatapult/dscp-matchmaker-api/blob/056c6779b313533f3dbb3eac3f3e72d53b886814/test/integration/order-demand.test.ts#L48 you can find test scenario here

n3op2: why 404 suggest that path is wrong, it suggest that item is not found 

```
In computer network communications, the HTTP 404, 404 not found, 404, 404 error, page not found or file not found error message is a hypertext transfer protocol (HTTP) standard response code, to indicate that the browser was able to communicate with a given server, but the server could not find what was requested.
```
Quote ^^^ 
@mattdean-digicatapult - sorry but when you get a moment could please settle this down?

dblane-digicatapult: Its 400.

n3op2: @dblane-digicatapult why it's 400? If we send a request with the UUID, knowing that it's in correct format, then what we do, we try to find an attachment, e.g. perform a db lookup for the given ID, the UUID/ID is in right format, so request itself is good? No?

n3op2: P.S still would like to hear why you wanted to use responses:
@dblane-digicatapult why 400 for item not found? (e.g. attachment) if the request URL contains UUID for the attachment that we do not have (server does not have)
@jonmattgray why 422 for the same reason?

jonmattgray (approved):

PR #40, opened 2023-03-29
Update transaction table

nitro-marky:
This migration replaces `token_type` -> `api_type` with options ['match2', 'order', 'capacity] and `transaction_type` options match the route that creates the transaction e.g. creation, proposal, accept

`token_id` column has been removed and all tests updated.

Screenshot 2023-03-29 at 12 21 13

### DIFF
--- a/src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
+++ b/src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
@@ -10,5 +10,13 @@ export async function up(knex: Knex): Promise<void> {
 }
 
 export async function down(knex: Knex): Promise<void> {
-  await knex.schema.dropTable('transaction')
+  return knex.schema.table('transaction', function (table) {
+    table.integer('token_id')
+    table.enu('token_type', ['DEMAND', 'MATCH2'], {
+      enumName: 'type',
+      useNative: true,
+    })
+    table.dropColumn('transaction_type')
+    table.dropColumn('api_type')
+  })
 }
--- a/src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
+++ b/src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
@@ -1,0 +1,14 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  return knex.schema.table('transaction', function (table) {
+    table.dropColumn('token_id')
+    table.dropColumn('token_type')
+    table.enu('api_type', ['match2', 'order', 'capacity'], { useNative: true, enumName: 'api_type' })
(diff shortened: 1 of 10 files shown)


src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
```diff
@@ -0,0 +1,14 @@
...
+  return knex.schema.table('transaction', function (table) {
+    table.dropColumn('token_id')
+    table.dropColumn('token_type')
+    table.enu('api_type', ['match2', 'order', 'capacity'], { useNative: true, enumName: 'api_type' })
+    table.enu('transaction_type', ['creation', 'proposal', 'accept'], { useNative: true, enumName: 'transaction_type' })
+  })
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.schema.dropTable('transaction')
```
jonmattgray, line 13: the down function should be a reverse of `up`to put the tablel back in the state it was in before e.g. https://github.com/digicatapult/inteli-api/blob/main/migrations/20220526113916_fix_tx_status_col.js

jonmattgray: you can readd `token_type` with 
```
    def
      .enum('token_type', ['DEMAND', 'MATCH2'], {
        enumName: 'type',
        useNative: true,
      })
      .notNullable()
```
but then you need to set a value for every row because it's `notNullable`. Which I think would mean looking at each rows `api_type` and mapping `match2` -> `MATCH2` and `order || capacity` -> `DEMAND`. But that means in the `up` you also need to do the reverse of `token_type` -> `api_type` for each row. And how do you decide if a `DEMAND` is an `order` or `capacity`.

It's complicated and also pointless anyway because we don't have any databases deployed with reall data

src/lib/db/migrations/20230329093154_transaction-table-modify-cols.ts
```diff
@@ -0,0 +1,22 @@
...
+    table.dropColumn('token_type')
+    table.enu('api_type', ['match2', 'order', 'capacity'], { useNative: true, enumName: 'api_type' })
+    table.enu('transaction_type', ['creation', 'proposal', 'accept'], { useNative: true, enumName: 'transaction_type' })
+  })
+}
+
+export async function down(knex: Knex): Promise<void> {
+  return knex.schema.table('transaction', function (table) {
+    table.integer('token_id')
+    table.enu('token_type', ['DEMAND', 'MATCH2'], {
```
jonmattgray, line 15: there's an issue in that `token_type` was `notNullable()` when originally created and the `down` isn't restoring that, so migrating up to this and down would delete the `token_type` value for existing transactions, but in the interest of moving on I'm going to approve

PR #42, opened 2023-03-30
L3-80 Get transaction

nitro-marky:
This adds:

GET /transaction API which lists all transactions. It also can take a status query param to filter transactions in a particular api status.

GET /transaction/{id} API which gets a specific transaction by id

Part One of the AC requirements.


test/integration/transaction.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        const response = await get(app, `/transaction/${nonExistentId}`)
+        expect(response.status).to.equal(404)
+      })
+
+      it('made-up transaction type - 422', async () => {
+        const response = await get(app, `/transaction?apiType=${'banana'}`)
+        expect(response.status).to.equal(422) 
+      })
+
+      it('it should not get all transactions - 200', async () => {
```
n3op2, line 128: it should not get all? maybe we should assert that it return an empty array?

test/integration/transaction.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  describe('sad path', () => {
+    it('non-existent transaction id - 404', async () => {
+        const response = await get(app, `/transaction/${nonExistentId}`)
+        expect(response.status).to.equal(404)
+      })
+
+      it('made-up transaction type - 422', async () => {
+        const response = await get(app, `/transaction?apiType=${'banana'}`)
+        expect(response.status).to.equal(422)
```
n3op2, line 125: similar here i think it also gives some validation details

test/integration/transaction.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      it('non-existent transaction type - 422', async () => {
+        const response = await get(app, `/transaction?apiType=${TransactionApiType.order}`)
+        expect(response.status).to.equal(422)
+      })
+  })
+
+  describe('sad path', () => {
+    it('non-existent transaction id - 404', async () => {
+        const response = await get(app, `/transaction/${nonExistentId}`)
+        expect(response.status).to.equal(404)
```
n3op2, line 120: can we also assert for error since it has entity name in it

src/lib/error-handler/index.ts
```diff
@@ -61,7 +61,7 @@ export const errorHandler = function errorHandler(
 ): ExResponse | void {
   if (err instanceof ValidateError) {
     logger.warn(`Handled Validation Error for ${req.path}: %s`, JSON.stringify(err.fields))
-    const { status, ...rest } = err
+    const { ...rest } = err
```
jonmattgray, line 64: this change is causing the attachment test to fail

nitro-marky: Yeah, I've taken that out

test/integration/transaction.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        expect(response.status).to.equal(200)
+        expect(response.body).to.deep.equal([
+            {
+                id: seededTransactionId,
+                api_type: TransactionApiType.capacity,
+                transaction_type: TransactionType.creation,
+                local_id: seededCapacityId,
+                state: TransactionState.submitted,
+                created_at: exampleDate,
+                updated_at: exampleDate,
```
jonmattgray, line 63: `expects` are failing because I changed `TransactionResponse` to use camel case keys to match rest of the repo. The tests is checking using `snake_case`

src/controllers/transaction/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Returns the details of all transactions.
+   * @summary List all transactions
+   * @Query apiType lists all transactions by that type
+   */
+  @Response<BadRequest>(400, 'Request was invalid')
+  @Response<NotFound>(404, 'Item not found')
+  @Get('/')
+  public async getAllTransactions(@Query() apiType?: TransactionType): Promise<TransactionResponse[]> {
```
jonmattgray, line 30: the query param says `apiType` but it's using `TransactionType` instead?

test/integration/capacity.test.ts
```diff
@@ -192,6 +192,7 @@ describe('capacity', () => {
     })
 
     it('non-existent Capacity ID should return nothing - 404', async () => {
+      await cleanup()
```
jonmattgray, line 195: what's the purpose of the `cleanup` here?

src/lib/error-handler/index.ts
jonmattgray: the change is still there?

src/controllers/transaction/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (apiType) {
+      const transactionsByType = await this.db.getTransactionsByType(apiType)
+      if (!transactionsByType) throw new NotFound('transactionsByType')
```
jonmattgray, line 33: if you give an invalid type to `getTransactionsByType` it will find no matches in the db and return `[]`. That won't cause this `if` to evaluate to `true` though so the `NotFound` doesn't get thrown. I think you'll need to check the `apiType` query is a valid value for the enum instead and throw an error if it's not

test/integration/capacity.test.ts
nitro-marky: To empty the db and check it returns nothing.

src/lib/error-handler/index.ts
n3op2: updated, please validate when you get a chance.

src/controllers/transaction/index.ts
n3op2: Yes, this got me for a little while, since I was looking at the tests when just started. If I would have done this earlier would have saved some time:), please confirm that I use the correct type.

test/integration/capacity.test.ts
n3op2: to reset the database so we can assert for getAll if no records are preset, this has been slightly updated

src/controllers/transaction/index.ts
n3op2: Yea. when you get a moment I think this might be resolved, will double check in a sec.

test/integration/capacity.test.ts
jonmattgray: the test is a `404` because there's no capacity with ID `nonExistentId`, there should be no need to reset the db since no seed adds a record for `nonExistentId`

src/controllers/transaction/index.ts
jonmattgray: I don't think the line `if (!transactionsByType) throw new NotFound('transactionsByType')` is needed? `transactionsByType` should never be falsy as the db query will return a `[]` if there's no match, and we can return `[]` rather than error

src/controllers/transaction/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  @Get('/')
+  public async getAllTransactions(@Query() apiType?: TransactionApiType): Promise<TransactionResponse[]> {
+    if (apiType) {
+      const transactionsByType = await this.db.getTransactionsByType(apiType)
+      if (!transactionsByType) throw new NotFound('transactionsByType')
+
+      return transactionsByType
+    } else {
+      const transactions = await this.db.getTransactions()
+      if (!transactions) throw new NotFound('transactions')
```
jonmattgray, line 38: similarly here, can remove. You can't `404 - not found` a `GET /`

test/integration/transaction.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      })
+  })
+
+  // TODO assert for limit (when set on the db)
+  it('returns all transactions', async () => {
+    const { status, body } = await get(app, '/transaction')
+
+    expect(status).to.equal(200)
+    expect(body.length).to.equal(5)
+    expect(body).to.deep.include.members([
```
jonmattgray, line 90: could change to `expect length > 0` and only check the the first transaction. Otherwise every time we change one of the seeded transactions (which get used for testing various different routes) we have to update this test.  We do have a ticket to use test fixtures instead though so will be resolved after that https://digicatapult.atlassian.net/browse/L3-100 and could ignore this

test/seeds/index.ts
```diff
@@ -166,7 +168,7 @@ export const seed = async () => {
 
   await db.transaction().insert([
     {
-      id: seededAcceptTransactionId,
+      id: seededTransactionId3,
```
jonmattgray, line 171: causing a test to fail

test/integration/transaction.test.ts
n3op2: well, i would debate that and argue that we should trust our seeds and if we retrieve all we should assert what we get not pick one. Due to their differences e.g. types, statuses and etc.

As a solution we could bring this number down or other solution for heavy comparison i use something like **snapshotting** - auto generated. so you do not need to updated manually 10 lines, you generate a new snapshot when you update properties or entities or...

src/controllers/transaction/index.ts
n3op2: lol:) did not look at the code, why it wasn't highlighted earlier?

test/integration/capacity.test.ts
n3op2: yea, sorry this is my bad, took a while to understand what was done.

test/seeds/index.ts
```diff
@@ -176,6 +178,18 @@ export const seed = async () => {
...
+      id: seededTransactionId4,
+      api_type: TransactionApiType.order,
+      transaction_type: TransactionType.creation,
+      local_id: seededCapacityId,
+      state: TransactionState.submitted,
+      created_at: exampleDate,
+      updated_at: exampleDate,
+    },
+  ])
+
```
n3op2, line 192: @jonmattgray I might be lost already, does this look all good?

jonmattgray (approved):

n3op2 (approved):

PR #45, opened 2023-03-31
Feature/l3 39

n3op2:
# What
- new POST controller for suibmitting transaction to the block chain
- get controllers for transactions and order
- integration tests (using temporarly mock, there is a story to address that)

# swagger
Screenshot 2023-03-31 at 14 39 02

### DIFF
--- a/test/integration/order.test.ts
+++ b/test/integration/order.test.ts
@@ -167,7 +167,6 @@ describe('order', () => {
 
     // check local transaction updates
     const [transaction] = await db.getTransaction(transactionId)
-    expect(transaction.state).to.equal(TransactionState.finalised)
     expect(transaction).to.contain({
       state: 'finalised',
       apiType: 'order',


src/controllers/order/index.ts
```diff
@@ -20,6 +38,66 @@ export class order extends Controller {
...
 
+  /**
+   * Returns the details of all order demands.
+   * @summary List all order demands
+   */
+  @Get('/')
+  public async getAll(): Promise<DemandResponse[]> {
+    const capacities = await this.db.getDemands(DemandSubtype.order)
+    const result = await Promise.all(
+      capacities.map(async (order: DemandResponse) => ({
```
jonmattgray, line 49: capacity from copy paste?

src/controllers/order/index.ts
```diff
@@ -20,6 +38,66 @@ export class order extends Controller {
...
+
+  /**
+   * @summary Get a order by ID
+   * @param orderId The order's identifier
+   */
+  @Response<NotFound>(404, 'Item not found')
+  @Get('{orderId}')
+  public async getCapacity(@Path() orderId: UUID): Promise<DemandResponse> {
+    const [order] = await this.db.getDemand(orderId)
+    if (!order) throw new NotFound('capacity')
```
jonmattgray, line 65: more `capacity`s

test/integration/order.test.ts
```diff
@@ -77,4 +81,38 @@ describe('order', () => {
...
+      state: 'finalised',
+      apiType: 'order',
+      transactionType: 'creation',
+    })
+
+    // check local capacity updates with token id
+    const [order] = await db.getDemand(orderId)
+    expect(order).to.contain({
+      id: orderId,
+      owner: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
```
jonmattgray, line 112: the convention is to return aliases for roles rather than addresses

test/integration/order.test.ts
```diff
@@ -77,4 +81,38 @@ describe('order', () => {
...
+    )
+    expect(state).to.equal(TransactionState.submitted)
+
+    // check local transaction updates
+    const [transaction] = await db.getTransaction(transactionId)
+    expect(transaction.state).to.equal(TransactionState.finalised)
+    expect(transaction).to.contain({
+      state: 'finalised',
+      apiType: 'order',
+      transactionType: 'creation',
```
jonmattgray, line 105: checking `state: finalised` twice? Also can use the enums (soon to be types!) for equality check

test/integration/order.test.ts
```diff
@@ -77,4 +81,38 @@ describe('order', () => {
...
+    })
+
+    // check local capacity updates with token id
```
jonmattgray, line 108: the test isn't checking `tokenId` at the moment

test/integration/order.test.ts
```diff
@@ -77,4 +81,38 @@ describe('order', () => {
...
+    const [order] = await db.getDemand(orderId)
+    expect(order).to.contain({
+      id: orderId,
+      owner: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
+      state: 'created',
+      subtype: 'order',
+      parametersAttachmentId,
+    })
+  })
 })
```
jonmattgray, line 118: could do with more tests to check the routes added `GET order/id` `GET order/` + sad path tests

n3op2: Could you please explain what scenarios are not covered?
I just did not cover for get `Validation Error`, also what about that tool? Also sad/happy path sounds weird, we need some scenarios, right?

A rule of thumb for  testing (one of) to test bad scenarios first and then everything else becomes a happy path without extra wrappers, and bad scenarios do trigger errors and we should assert error and also validate that other method have not been called after the error.

Basically, I can walk you through and share some of mine knowledge when it comes to TDD

test/integration/order.test.ts
n3op2: I'm not going to use enums for values, sorry but I can not allow myself to do that.

src/controllers/order/index.ts
n3op2: updated

n3op2: Looks like one of the unstagged changes. 
Screenshot 2023-04-03 at 13 19 26

Please confirm

src/controllers/order/index.ts
n3op2: same as above, did not pull, did not push, please confirm

src/controllers/order/index.ts
```diff
@@ -20,6 +38,66 @@ export class order extends Controller {
...
+      ...order,
+      alias: await getMemberByAddress(order.owner),
+    }
+  }
+
+  /**
+   * A member creates the capacity {capacityId} on-chain. The capacity is now viewable to other members.
+   * @summary Create a new capacity demand on-chain
+   * @param orderId The capacity's identifier
+   */
```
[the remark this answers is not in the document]
n3op2: same as above. please confirm

test/integration/order.test.ts
n3op2: Updated

test/integration/order.test.ts
jonmattgray: you're using enum as value with `expect(transaction.state).to.equal(TransactionState.finalised)` so why the inconsistency?

test/integration/order.test.ts
jonmattgray: still address?

test/integration/order.test.ts
jonmattgray: I see more tests have been added since I commented. There's no `200` test for `GET order/id` nor one testing the `400` on order creation `if (order.state !== DemandState.created)`

src/controllers/order/index.ts
```diff
@@ -20,6 +38,66 @@ export class order extends Controller {
...
+  @Response<NotFound>(404, 'Item not found')
+  @Get('{orderId}')
+  public async getById(@Path() orderId: UUID): Promise<DemandResponse> {
+    const [order] = await this.db.getDemand(orderId)
+    if (!order) throw new NotFound('order')
+
+    return {
+      ...order,
+      alias: await getMemberByAddress(order.owner),
```
jonmattgray, line 69: we replace the `owner` key's value with the alias rather than assigning to a new key, same issue on the `GetAll`. `create` in this file returns `owner` correctly

test/integration/order.test.ts
n3op2: yea, totally forgot that added get controllers, only for demo. This has been updated

test/integration/order.test.ts
```diff
@@ -122,10 +156,21 @@ describe('order', () => {
     const [order] = await db.getDemand(orderId)
     expect(order).to.contain({
       id: orderId,
-      owner: 'test-self',
```
jonmattgray: why remove?

test/integration/order.test.ts
jonmattgray: still not seeing a test for `if (order.state !== DemandState.created)`?

test/integration/order.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    })
+
+    const [order] = await db.getDemand(orderId)
+    expect(transaction.localId).to.equal(orderId)
+    expect(order).to.contain({
+      id: orderId,
+      state: 'created',
+      subtype: 'order',
+      parametersAttachmentId,
+    })
```
jonmattgray, line 163: would be good to check token ids on the order too

src/controllers/order/index.ts
```diff
@@ -20,6 +37,83 @@ export class order extends Controller {
...
+  }
+
+  /**
+   * @summary Get a order creation transaction by ID
+   * @param orderId The order's identifier
+   * @param creationId The order's creation ID
+   */
+  @Response<NotFound>(404, 'Item not found.')
+  @SuccessResponse('200')
+  @Get('{orderId}/creation/{creationId}')
```
jonmattgray, line 79: there's the get all creation transactions route as well `@Get('{orderId}/creation/'`. Depends if you want to do in this PR

n3op2: Yea, I did not want to do transactions, but since I started, will put that in along with a test.

test/integration/order.test.ts
jonmattgray: to show it's been put on chain and its token ID has updated locally

test/integration/order.test.ts
n3op2: must have forgotten this one. Updated. thanks.

test/integration/order.test.ts
n3op2: done

test/integration/order.test.ts
mattdean-digicatapult: Um I think the point is ln 101 is literally contained within the assertion on ln 102. That said this is a nothing burger and this shouldn't prevent this merging.

@n3op2 I'm not sure what you're proposing here other than making a general gripe? If there's a point to make then fine but that's not how this is coming across

n3op2: @mattdean-digicatapult not sure if I understood you. I just found this picky.

n3op2: @mattdean-digicatapult yes, line has been removed, no need to double assert:) - useless

jonmattgray (approved):

PR #46, opened 2023-03-31
Chore/processed blocks table

mattdean-digicatapult:
Creates structures in place for the blockchain indexer. Todo:

- [x] db migration script
- [x] env handling to enable indexer
- [x] basic indexer structure

### DIFF
--- a/src/lib/db/migrations/20230331084321_processed-blocks-table.ts
+++ b/src/lib/db/migrations/20230331084321_processed-blocks-table.ts
@@ -1,0 +1,20 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable('processed_blocks', (def) => {
+    def.specificType('hash', 'CHAR(64)').notNullable()
+    def.bigInteger('height').unsigned().notNullable().unique()
+    def.specificType('parent', 'CHAR(64)').notNullable()
+    def.datetime('created_at').notNullable().defaultTo(knex.fn.now())
+
+    def.primary(['hash'])
+  })
+
+  await knex.schema.alterTable('processed_blocks', (def) => {
+    def.foreign('parent', 'fk_processed_blocks_parent_hash').references('hash').inTable('processed_blocks')
+  })
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.schema.dropTable('processed_blocks')
+}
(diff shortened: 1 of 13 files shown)


src/lib/services/blockchainWatcher.ts
```diff
@@ -1,23 +1,27 @@
...
-const typeTableMap = {
-  [TokenType.DEMAND]: 'demand',
-  [TokenType.MATCH2]: 'match2',
+const typeToTable = (tokenType: TOKEN_TYPE) => {
+  switch (tokenType) {
+    case 'DEMAND':
+      return 'demand'
+    case 'MATCH2':
+      return 'match2'
+  }
```
n3op2, line 16: const typeToTable = (tokenType: TOKEN_TYPE) => tokenType.toLowerCase()? why switch? Just out of interest

n3op2 (approved):

src/lib/db/migrations/20230331084321_processed-blocks-table.ts
```diff
@@ -0,0 +1,20 @@
...
+    def.specificType('hash', 'CHAR(64)').notNullable()
+    def.bigInteger('height').unsigned().notNullable().unique()
+    def.specificType('parent', 'CHAR(64)').notNullable()
+    def.datetime('created_at').notNullable().defaultTo(knex.fn.now())
+
+    def.primary(['hash'])
+  })
+
+  await knex.schema.alterTable('processed_blocks', (def) => {
+    def.foreign('parent', 'fk_processed_blocks_parent_hash').references('hash').inTable('processed_blocks')
```
jonmattgray, line 14: how would the first insert work? Seems like a chicken/egg thing, you need an existing row with a hash to set as parent, and that row needs a row for its parent and so on...

jonmattgray: since the columns are `NotNullable`

mattdean-digicatapult: A row can reference itself

src/lib/services/blockchainWatcher.ts
mattdean-digicatapult: it forces the correct type coertion. I tried to do this with an object map but typescript wasn't happy with that

src/lib/indexer/index.ts
```diff
@@ -0,0 +1,48 @@
+import { Logger } from 'pino'
```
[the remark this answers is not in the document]
mattdean-digicatapult: I hate it to be honest. I'm really not a fan of OOP and class structures but I'm trying to be consistent with the style of the repo

PR #50, opened 2023-04-04
Replace dscp-api with own runProcess

jonmattgray:
- Replaces `dscp-api` with a local `runProcess` that uses `polkadot` + `ipfs` 
- Integration tests put tokens on chain rather than mocking `dscp-api` responses


src/controllers/attachment/index.ts
```diff
@@ -102,11 +103,14 @@ export class attachment extends Controller {
 
     if (!req.body && !file) throw new BadRequest('nothing to upload')
 
-    const [{ id, filename, binary_blob, created_at }]: any[] = await this.db
+    const buffer = Buffer.from(file?.buffer || JSON.stringify(req.body))
+    if (buffer.length > env.FILE_UPLOAD_MAX_SIZE) throw new BadRequest(`Over size limit: ${env.FILE_UPLOAD_MAX_SIZE}`)
```
jonmattgray, line 107: seeing `FILE_UPLOAD_MAX_SIZE` in `dscp-api` made me realise we weren't limiting it in this API

src/lib/db/index.ts
```diff
@@ -86,12 +86,12 @@ export default class Database {
     return this.db().demand().select(demandColumns).where({ id })
   }
 
-  getDemandWithAttachment = async (capacityId: UUID, subtype: DemandSubtype) => {
+  getDemandWithAttachment = async (id: UUID) => {
     return this.db()
       .demand()
       .join('attachment', 'demand.parameters_attachment_id', 'attachment.id')
       .select()
-      .where({ 'demand.id': capacityId, subtype })
```
jonmattgray, line 94: temporary change until we have `POST order/id/creation`, otherwise I put an order on chain in order to test match2s on chain

jonmattgray: it means that someone can create an `order` locally then `POST capacity/id/creation` and it still works...

src/lib/payload.ts
```diff
@@ -2,8 +2,25 @@ import { Match2Payload, Match2Response, Match2State } from '../models/match2'
...
+  roles: Record<string, string>
+  metadata: Metadata
+}
+
+export interface MetadataFile {
+  blob: Blob
+  filename: string
+}
+
+export type Metadata = Record<string, { type: string; value: string | MetadataFile | number }>
```
jonmattgray, line 21: not sure where these types should live?

test/integration/match2.test.ts
```diff
@@ -170,53 +227,6 @@ describe('match2', () => {
...
-      // submit to chain
-      const response = await post(app, `/match2/${seededMatch2OrderNotOwnedId}/accept`, {})
-      expect(response.status).to.equal(201)
-
-      // check local entities update with token id
-      const [match2] = await db.getMatch2(seededMatch2OrderNotOwnedId)
-      expect(match2.latestTokenId).to.equal(match2AcceptMockTokenId)
-      expect(match2.state).to.equal(Match2State.acceptedB)
-      expect(match2.originalTokenId).to.equal(seededMatch2TokenId)
-    })
```
jonmattgray, line 218: this is the test scenario that will require mocking `ChainNode` because the `demand-create` process flow makes it impossible to put a demand on chain that's not owned by the current user

src/lib/chainNode.ts
```diff
@@ -48,4 +70,126 @@ export default class ChainNode {
...
+          const { dispatchError, status } = result
+
+          if (dispatchError) {
+            if (dispatchError.isModule) {
+              const decoded = this.api.registry.findMetaError(dispatchError.asModule)
+              reject(new HttpResponse({ message: `Node dispatch error: ${decoded.name}` }))
+            } else {
+              reject(Error(dispatchError.toString()))
+            }
+          }
```
jonmattgray, line 129: what do you think of this error handling @mattdean-digicatapult, we havent used `dispatchError` before I think?

src/lib/services/ipfs.ts
mattdean-digicatapult, line 1: it may be worth making this a class for a consistent structure with the rest of the repo

src/lib/db/index.ts
jonmattgray: https://github.com/digicatapult/dscp-matchmaker-api/pull/45 will add it if that goes in first

src/controllers/attachment/index.ts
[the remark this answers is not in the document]
jonmattgray: not immediately seeing how to set multer config when it's part of `tsoa`. I'll remove setting a size limit from this PR since it's not really part of what I'm doing + add a new ticket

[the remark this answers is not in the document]
jonmattgray: https://digicatapult.atlassian.net/browse/L3-117

src/lib/chainNode.ts
jonmattgray: I think we can use `dispatchError` instead of filtering the events as we have been:
```
            const errors = result.events
              .filter(({ event: { method } }) => method === 'ExtrinsicFailed')
              .map(({ event: { data } }) => data[0])
              
             if (errors.length > 0) {
              reject(ProcessExtrinsicError(errors[0]))
            }
```

e.g. for a `ProcessInvalid`, `dispatchError === errors[0]`

jonmattgray: I'll change L127 to a `HttpResponse` error too

src/lib/chainNode.ts
```diff
@@ -1,23 +1,45 @@
...
+
 export default class ChainNode {
   private provider: WsProvider
   private api: ApiPromise
-  // private keyring: Keyring
+  private keyring: Keyring
   private logger: Logger
+  private roles: RoleEnum[]
 
   constructor({ host, port, logger }: NodeCtorConfig) {
```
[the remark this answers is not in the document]
jonmattgray: done

src/lib/services/ipfs.ts
jonmattgray: done

src/lib/chainNode.ts
jonmattgray: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-get-the-decoded-enum-for-an-extrinsicfailed-event

jonmattgray:
```
As of the @polkadot/api 2.3.1 additional result fields are exposed. Firstly there is dispatchInfo: DispatchInfo which occurs in both ExtrinsicSuccess & ExtrinsicFailed events. Additionally, on failures the dispatchError: DispatchError is exposed. With this in mind, the above can be simplified to be -
```

dblane-digicatapult (changes requested):

src/env.ts
```diff
@@ -17,9 +17,10 @@ export default envalid.cleanEnv(process.env, {
   DB_NAME: envalid.str({ default: 'dscp-matchmaker-api' }),
   IDENTITY_SERVICE_HOST: envalid.host({ devDefault: 'localhost' }),
   IDENTITY_SERVICE_PORT: envalid.port({ devDefault: 3002 }),
```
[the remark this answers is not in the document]
jonmattgray: done

src/env.ts
```diff
@@ -17,9 +17,10 @@ export default envalid.cleanEnv(process.env, {
...
-  DSCP_API_HOST: envalid.host({ devDefault: 'localhost' }),
-  DSCP_API_PORT: envalid.port({ devDefault: 3001 }),
   NODE_HOST: envalid.host({ default: 'localhost' }),
   NODE_PORT: envalid.port({ default: 9944 }),
   ENABLE_INDEXER: envalid.bool({ default: false }),
+  USER_URI: envalid.str({ devDefault: '//Alice' }),
+  IPFS_HOST: envalid.host({ devDefault: 'localhost' }),
+  IPFS_PORT: envalid.port({ devDefault: 5001, default: 15001 }),
```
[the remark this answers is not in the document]
jonmattgray: done

dblane-digicatapult (approved):

PR #51, opened 2023-04-06
Feature/indexer part3

mattdean-digicatapult:
Adds tests for the indexer and starts to build up the structure of how database inserts will happen in a transaction


src/lib/db/index.ts
```diff
@@ -152,12 +154,26 @@ export default class Database {
...
+  }
+
+  withTransaction = (update: (db: Database) => Promise<void>) => {
+    return this.client.transaction(async (trx) => {
+      const decorated: Database = {
+        ...this,
+        client: trx,
+      }
+      await update(decorated)
+    })
```
mattdean-digicatapult, line 177: Note this is a little dicey. I would prefer client to be dependency injected so I could create a new `Database` here properly with the decorated client. That said it works quite gracefully becasue of how Typescript handles private class properties

src/lib/indexer/__tests__/fixtures/db.ts
```diff
@@ -0,0 +1,44 @@
...
+  return {
+    getLastProcessedBlock: getMock,
+    insertProcessedBlock,
+    withTransaction: sinon.spy(async function (fn: (db: Database) => Promise<void>) {
+      await fn({
+        insertProcessedBlock,
+      } as unknown as Database)
+    }),
+  } as unknown as Database
+}
```
mattdean-digicatapult, line 44: It's worth saying that I'm torn on whether this is a good mocking approach in this instance. This is essentially a working mock and so I'm relying on the behaviour of the working mock against other assertions rather than asserting directly that the db was updated correctly. Not ideal but the mock is very simple so I think it's probably ok

src/lib/indexer/index.ts
```diff
@@ -1,27 +1,37 @@
 import { Logger } from 'pino'
 
-import type Database from '../db/index.js'
-import ChainNode from '../chainNode.js'
+import type Database from '../db/index'
+import ChainNode from '../chainNode'
+
+export type ChangeSet = Record<string, never>
```
mattdean-digicatapult, line 6: Hopefully it's obvious but the intent is to update this type to include the different types of upserts that will need to be made agains tthe database. Blocks will then be processed into a set of changes that can then all be applied in a single tx

jonmattgray (approved):

PR #56, opened 2023-04-11
Add Status parameter to GET Transactions

nitro-marky:
As per the title, as another optional parameter to GET Transactions, so it now takes `apiType` and `status`.


src/controllers/transaction/index.ts
```diff
@@ -25,9 +25,25 @@ export class TransactionController extends Controller {
...
+
+    // Checks if both are truthy and that both are valid in the enum
+    // Status has to be checked to stop 'Type 'undefined' is not assignable to type 'TransactionState' error
+    if (
+      apiType &&
+      status &&
+      Object.values(TransactionApiType).includes(apiType as TransactionApiType) &&
+      Object.values(TransactionState).includes(status as TransactionState)
+    )
+      return await this.db.getTransactionsByStateAndType(status, apiType)
```
jonmattgray, line 46: rather than checking `apiType !=undefined` and `status !=undefined` etc. I think you might be able to always pass both by using optional parameters for `getTransactionsByStateAndType`. If it's not passed a value use `*` (meaning any) for the where clause. e.g. if `status=submitted` and `apiType=undefined`, then the db query is `where({ state: 'submitted', api_type: '*' }).`

jonmattgray: and it would mean the only db query needed is `getTransactionsByStateAndType`

jonmattgray (approved):

PR #61, opened 2023-04-13
Replace all enums with string unions

jonmattgray:
Replace all enums with string unions because enums in typescript are bad/redundant/cause issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@digicatapult/dscp-matchmaker-api",
-  "version": "0.6.4",
+  "version": "0.6.5",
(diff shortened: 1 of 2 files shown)


src/lib/chainNode.ts
```diff
@@ -161,6 +161,7 @@ export default class ChainNode {
           unsub = res
         })
         .catch((err) => {
+          transactionDbUpdate('failed')
```
jonmattgray, line 164: sneaking this in, should have been in #58

nitro-marky (approved):

mattdean-digicatapult (approved):

PR #63, opened 2023-04-18
Sets the max upload limit

nitro-marky:
This sets a limit of 100mb as an upload limit for the entire tsoa aplcation to ensure nothing outrageous gets uploaded.

### DIFF
--- a/test/integration/offchain/attachment.test.ts
+++ b/test/integration/offchain/attachment.test.ts
@@ -13,7 +13,7 @@ describe('attachment', async () => {
   const size = 100
   const blobData = 'a'.repeat(size)
   const filename = 'test.pdf'
-  const overSize = 10242880
+  const overSize = 115343360
   const overSizeBlobData = 'a'.repeat(overSize)
   const jsonData = { key: 'it', filename: 'JSON attachment it' }
   let app: Express
@@ -125,7 +125,7 @@ describe('attachment', async () => {
   })
 
 
-    it('Doesn`t upload files if more than 5mb', async () => {
+    it('Doesn`t upload files if more than 100mb', async () => {
     const uploadRes = await postFile(app, '/attachment', Buffer.from(overSizeBlobData), 'json')
     const { status  } = await get(app, `/attachment/${uploadRes.body.id}`)
     console.log(status)
--- a/test/integration/offchain/attachment.test.ts
+++ b/test/integration/offchain/attachment.test.ts
@@ -9,10 +9,12 @@ import Database from '../../../src/lib/db'
 
 const db = new Database().db()
 
-describe('attachment', () => {
+describe('attachment', async () => {
   const size = 100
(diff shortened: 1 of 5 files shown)


test/integration/offchain/attachment.test.ts
```diff
@@ -121,4 +123,14 @@ describe('attachment', () => {
...
   })
+
+
+    it('Doesn`t upload files if more than 100mb', async () => {
+    const uploadRes = await postFile(app, '/attachment', Buffer.from(overSizeBlobData), 'json')
+    const { status  } = await get(app, `/attachment/${uploadRes.body.id}`)
+    console.log(status)
+    expect(status).to.equal(422)
+
+    })
```
mattdean-digicatapult, line 134: Surprised the linter didn't catch this but can you just fix the indentation here

tsoa.json
```diff
@@ -4,6 +4,11 @@
   "controllerPathGlobs": [
     "src/controllers/**/*.ts"
   ],
+  "multerOpts": {
+    "limits": {
+      "fileSize": 104857600
+    }
+  },
```
n3op2, line 11: boom!

test/integration/offchain/attachment.test.ts
```diff
@@ -9,10 +9,12 @@ import Database from '../../../src/lib/db'
 
 const db = new Database().db()
 
-describe('attachment', () => {
+describe('attachment', async () => {
```
n3op2, line 12: why is this is async? is it because of .repeat?

test/integration/offchain/attachment.test.ts
```diff
@@ -121,4 +123,12 @@ describe('attachment', () => {
       'content-disposition': 'attachment; filename="json"',
     })
   })
+
+
+    it('Doesn`t upload files if more than 100mb', async () => {
```
n3op2, line 128: nice one for the test to back this up


can you please also assert for body?

test/integration/offchain/attachment.test.ts
```diff
@@ -121,4 +123,12 @@ describe('attachment', () => {
...
+      const uploadRes = await postFile(app, '/attachment', Buffer.from(overSizeBlobData), 'json')
+      const { status  } = await get(app, `/attachment/${uploadRes.body.id}`)
```
n3op2, line 130: FYI: const { status, body ....

n3op2 (commented):
nice one, but please can you assert for body to validate error message as well?

test/integration/offchain/attachment.test.ts
nitro-marky: But surely with this I only need the response to prove that the file has not been uploaded?

mattdean-digicatapult (approved):

n3op2 (changes requested):

PR #64, opened 2023-04-18
Feat/block handler part 1

mattdean-digicatapult:
Implements some basic structures around block handling. Still missing are:

* handleEvent doesn't actually process the event and is currently untested. This will change in the next PR
* the event processors are not implemented but the basic type structure is in place. Again next PR


src/lib/chainNode.ts
```diff
@@ -9,6 +10,8 @@ import { HttpResponse } from './error-handler'
 import Ipfs from './ipfs'
 import type { Payload, Output, Metadata, MetadataFile } from './payload'
 
+const processRanTopic = blake2AsHex('simpleNFT.ProcessRan')
```
jonmattgray, line 13: utxoNFT

jonmattgray (approved):

PR #67, opened 2023-04-19
Chore/ipfs upload on attachment

mattdean-digicatapult:
We need to upload on attachment creation rather than on submission so that the db doesn't need to store binary data as this will be unavailabe to the block indexer. Also note size has been made optional to reflect that the indexer will not have this information either but instead it is updated on fetching the resource. I will also try to get this information oportunistically in the indexer but if we don't have it null is better than the indexer failing.

### DIFF
--- a/src/controllers/attachment/index.ts
+++ b/src/controllers/attachment/index.ts
@@ -22,6 +22,8 @@ import type { Attachment } from '../../models'
 import { BadRequest, NotFound } from '../../lib/error-handler'
 import { Readable } from 'node:stream'
 import type { UUID } from '../../models/uuid'
+import Ipfs from '../../lib/ipfs'
+import env from '../../env'
 
 const parseAccept = (acceptHeader: string) =>
   acceptHeader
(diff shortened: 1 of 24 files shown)


src/controllers/attachment/index.ts
```diff
@@ -84,12 +96,11 @@ export class attachment extends Controller {
   public async get(): Promise<Attachment[]> {
     this.log.debug('retrieving all attachment')
 
-    const attachments: Attachment[] = await this.db.attachment()
+    const attachments: DbAttachment[] = await this.db.attachment()
     return attachments.map(
-      ({ binary_blob, created_at, ...rest }: any): Attachment => ({
+      ({ created_at, ipfs_hash, ...rest }): Attachment => ({
```
jonmattgray, line 101: add an eslint comment for the deliberate unused variable?

mattdean-digicatapult: This is such a common pattern I'm going to suggest updating our eslint rules

jonmattgray (approved):

PR #69, opened 2023-04-20
API Status Watcher

nitro-marky:
The first of API, IPFS and Identity watchers to be added.

Example output as of 16/5/23:

```
{
    "status": "down",
    "version": "1.2.6",
    "details": {
        "api": {
            "status": "ok",
            "detail": {
                "chain": "Development",
                "runtime": {
                    "name": "dscp",
                    "versions": {
                        "spec": 802,
                        "impl": 1,
                        "authoring": 1,
                        "transaction": 1
                    }
                }
            }
        },
        "ipfs": {
            "status": "down",
            "detail": {
                "message": "Error getting status from IPFS node"
            }
        }
    }
}
```


src/env.ts
```diff
@@ -23,4 +23,7 @@ export default envalid.cleanEnv(process.env, {
   USER_URI: envalid.str({ devDefault: '//Alice' }),
   IPFS_HOST: envalid.host({ devDefault: 'localhost' }),
   IPFS_PORT: envalid.port({ default: 5001 }),
+  SUBSTRATE_STATUS_POLL_PERIOD_MS: envalid.num({ default: 10 * 1000 }),
```
n3op2, line 26: just wonder maybe we could generalise this one and call something like `WATCHER_POLL_PERIOD_MS` - first thought pretty sure it could have some more relevant name.

Mainly because we will be polling other services pretty much save interval. If we won't then we would get some discrepancies or would need to define multiple env vars

src/util/statusPoll.ts
```diff
@@ -0,0 +1,109 @@
+export const serviceState = {
```
n3op2, line 1: please treat `/lib` as 'util' mentioned in another comment, would be great if we could move `ServiceWatcher` related stuff into `src/lib/ServiceWatcher`

src/util/substrateApi.ts
```diff
@@ -0,0 +1,27 @@
...
+  logger.warn(`Disconnected from substrate node at ${NODE_HOST}:${NODE_PORT}`)
+})
+
+api.on('connected', () => {
+  logger.info(`Connected to substrate node at ${NODE_HOST}:${NODE_PORT}`)
+})
+
+api.on('error', (err) => {
+  logger.error(`Error from substrate node connection. Error was ${err.message || JSON.stringify(err)}`)
+})
```
n3op2, line 24: I think this is duplicate of https://github.com/digicatapult/dscp-matchmaker-api/blob/main/src/lib/chainNode.ts

To be fair I'm a little confused, why do you need this?

n3op2 (commented):
Looks like a lot has been implemented. However, left a few questions. Also could you please add some requests/responses to the description of this PR, please.

src/env.ts
```diff
@@ -23,4 +23,7 @@ export default envalid.cleanEnv(process.env, {
   USER_URI: envalid.str({ devDefault: '//Alice' }),
   IPFS_HOST: envalid.host({ devDefault: 'localhost' }),
   IPFS_PORT: envalid.port({ default: 5001 }),
+  WATCHER_POLL_PERIOD_MS: envalid.num({ default: 10 * 1000 }),
+  WATCHER_TIMEOUT_MS: envalid.num({ default: 2 * 1000 }),
+  API_VERSION: envalid.str({ default: process.env.npm_package_version }),
```
jonmattgray, line 28: we probably don't need the API_VERSION env?

jonmattgray: in fact doesn't look like it's being used anywhere

src/server.ts
```diff
@@ -4,17 +4,42 @@ import { setup, serve } from 'swagger-ui-express'
...
+  const serviceStatusStrings = {
+    [serviceState.UP]: 'ok',
+    [serviceState.DOWN]: 'down',
+    [serviceState.ERROR]: 'error',
+  }
   app.use(urlencoded({ extended: true }))
   app.use(json())
   app.use(cors())
 
+  app.get('/health', async (_req, res) => {
```
jonmattgray, line 24: there's a health route already defined which you'll want to use instead of putting it in `server.ts`. Putting it in `/controllers` means tsoa will build the documentation for swagger https://github.com/digicatapult/dscp-matchmaker-api/blob/main/src/controllers/health/index.ts

package.json
```diff
@@ -75,6 +75,7 @@
     "express": "^4.18.2",
     "knex": "^2.4.2",
     "multer": "^1.4.5-lts.1",
+    "node-fetch": "^3.3.1",
```
jonmattgray, line 78: now we're on node 18, `fetch` is built-in, you don't need the package
https://dev.to/cloudx/nodejs-18-fetch-api-test-runner-module-and-more-2ckg

src/lib/ServiceWatcher/ipfsStatus.ts
```diff
@@ -0,0 +1,49 @@
+import fetch from 'node-fetch'
+import { startStatusHandler, serviceState } from '../util/statusPoll'
+import env from '../../env'
+
+const { WATCHER_POLL_PERIOD_MS, WATCHER_TIMEOUT_MS, IPFS_HOST, IPFS_PORT } = env
+
+const versionURL = `http://${IPFS_HOST}:${IPFS_PORT}/api/v0/version`
+const peersURL = `http://${IPFS_HOST}:${IPFS_PORT}/api/v0/swarm/peers`
+
+const getStatus = async () => {
```
jonmattgray, line 10: Could move this function into the existing `ipfs.ts` class since the ipfs config is already there and it keeps all calls to ipfs in the same place. Then the service watcher constructor can include an instance of the `ipfs` class.

Same thing with `getStatus` for substrate, that could live in the `chainNode.ts` class and service watcher has an instance of `chainNode`.

package.json
nitro-marky: I kept getting a depcheck error that it was missing, even though it wasn't and I removed `node_modules` and ran `npm i` and still got the error

jonmattgray: you import it `import fetch from 'node-fetch'` in `ipfsStatus.ts`. Can delete that

jonmattgray: `fetch` works without an import in the file

Dockerfile
```diff
@@ -25,4 +25,4 @@ RUN npm ci --production
 COPY --from=builder /dscp-matchmaker-api/build .
 
 EXPOSE 80
-CMD [ "node", "./index.js" ]
\ No newline at end of file
+CMD [ "npm", "./index.js" ]
```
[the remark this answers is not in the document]
jonmattgray: I have a feeling this should be `CMD ["npm", "start"]` but hard to tell without testing it

src/lib/ipfs.ts
```diff
@@ -14,12 +15,16 @@ export default class Ipfs {
...
+  private versionURL: string
+  private peersURL: string
 
   constructor({ host, port, logger }: { host: string; port: number; logger: Logger }) {
     this.addUrl = `http://${host}:${port}/api/v0/add?cid-version=0&wrap-with-directory=true`
     this.dirUrl = (dirHash) => `http://${host}:${port}/api/v0/ls?arg=${dirHash}`
     this.fileUrl = (fileHash) => `http://${host}:${port}/api/v0/cat?arg=${fileHash}`
     this.logger = logger.child({ module: 'ipfs' })
+    this.versionURL = `http://${env.IPFS_HOST}:${env.IPFS_PORT}/api/v0/version`
+    this.peersURL = `http://${env.IPFS_HOST}:${env.IPFS_PORT}/api/v0/swarm/peers`
```
jonmattgray, line 27: `IPFS_HOST` and `IPFS_PORT` are already passed to the constructor as `host` and `port`

src/lib/util/statusPoll.ts
```diff
@@ -0,0 +1,109 @@
+export const serviceState = {
```
jonmattgray, line 1: this file could live in `ServiceWatcher` rather than `util`?

test/integration/offchain/healthcheck.test.ts
```diff
@@ -13,7 +13,6 @@ describe('health check', () => {
 
   it('should return 200', async () => {
     const response = await get(app, '/health')
```
jonmattgray, line 15: there needs to be more to the `200` test, asserting that the body contains what we expect. Also a test for if each service is down and what the body look like then. You'll need to mock the services to make them appear 'down'. e.g. https://github.com/digicatapult/dscp-api/blob/639934592c2e5b518ac14079bcccdd490e51f4e1/test/integration/routes.test.js#L97

src/controllers/health/index.ts
```diff
@@ -1,22 +1,46 @@
...
-      req_id: req.req_id,
-    })
+    const serviceStatusStrings = {
+      [serviceState.UP]: 'ok',
+      [serviceState.DOWN]: 'down',
+      [serviceState.ERROR]: 'error',
+    }
+
+    const statusHandler = await startStatusHandlers()
+    const status = statusHandler.status
```
jonmattgray, line 26: in `dscp-api` the `/health` endpoint returns a code based on `serviceState.UP ? 200 : 503`. In this repo with `tsoa` you'll need to throw a custom `ServiceUnavailable` error like how we throw `NotFound` and `BadRequest` in other controllers. It'll be a bit different to the `NotFound/BadRequest` errors because we send more than just a `message` in the body, I expect you'll need some new code in the error handler func to make it construct the correct response, like how we do for `422`s https://github.com/digicatapult/dscp-matchmaker-api/blob/main/src/lib/error-handler/index.ts#L62-L70

Dockerfile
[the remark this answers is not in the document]
jonmattgray: how to test:
```
docker build . -t somename
docker run somename
```
then look at the logs to see it's starting correctly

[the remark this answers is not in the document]
jonmattgray: you can pass envs like
`docker run -e DB_HOST='localhost' somename`

src/lib/ServiceWatcher/index.ts
```diff
@@ -0,0 +1,10 @@
+import startApiStatus from './apiStatus'
+import { buildCombinedHandler } from '../util/statusPoll'
+
+export const startStatusHandlers = async () => {
+  const handlers = new Map()
+  const [apiStatus] = await Promise.all([startApiStatus()])
+  handlers.set('api', apiStatus)
```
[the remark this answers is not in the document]
jonmattgray: sorry, I missed that

src/lib/ipfs.ts
```diff
@@ -14,12 +14,16 @@ export default class Ipfs {
...
+  private versionURL: string
+  private peersURL: string
 
   constructor({ host, port, logger }: { host: string; port: number; logger: Logger }) {
     this.addUrl = `http://${host}:${port}/api/v0/add?cid-version=0&wrap-with-directory=true`
     this.dirUrl = (dirHash) => `http://${host}:${port}/api/v0/ls?arg=${dirHash}`
     this.fileUrl = (fileHash) => `http://${host}:${port}/api/v0/cat?arg=${fileHash}`
     this.logger = logger.child({ module: 'ipfs' })
+    this.versionURL = `http://${host}:${port}/api/v0/version`
+    this.peersURL = `http://${host}:${port}/api/v0/swarm/peers`
```
mattdean-digicatapult, line 26: you'll need to mock these for the test @nitro-marky

test/helper/mockHealth.ts
```diff
@@ -0,0 +1,87 @@
...
+  let originalDispatcher: Dispatcher
+  let mockAgent: MockAgent
+  beforeEach(async function () {
+    originalDispatcher = getGlobalDispatcher()
+    mockAgent = new MockAgent()
+    setGlobalDispatcher(mockAgent)
+    const mockIdentity = mockAgent.get(`http://${env.IPFS_HOST}:${env.IPFS_PORT}`)
+    mockIdentity
+      .intercept({
+        path: `http://${env.IPFS_HOST}:${env.IPFS_PORT}/api/v0/version`,
```
mattdean-digicatapult, line 19: I think there's duplicate information here. Should this not be `/api/v0/version` because the host is already set on ln 16

src/lib/error-handler/index.ts
```diff
@@ -1,5 +1,7 @@
 import { Response as ExResponse, Request as ExRequest, NextFunction } from 'express'
 import { ValidateError } from 'tsoa'
+import { Health } from '../../models'
+//import { Health } from '../../models'
```
jonmattgray, line 4: can delete

test/helper/healthHelper.ts
```diff
@@ -0,0 +1,63 @@
+export const responses = {
+  ok: (dscpRuntimeVersion: any, ipfsVersion: any) => ({
```
jonmattgray, line 2: these `any`s are presumably `string`?

src/env.ts
```diff
@@ -23,4 +23,6 @@ export default envalid.cleanEnv(process.env, {
   USER_URI: envalid.str({ devDefault: '//Alice' }),
   IPFS_HOST: envalid.host({ devDefault: 'localhost' }),
   IPFS_PORT: envalid.port({ default: 5001 }),
+  WATCHER_POLL_PERIOD_MS: envalid.num({ default: 10 * 1000 }),
+  WATCHER_TIMEOUT_MS: envalid.num({ default: 2 * 1000 }),
```
jonmattgray, line 27: need adding to README

test/helper/healthHelper.ts
mattdean-digicatapult: I think they are a mixture of string and number. Will update

jonmattgray (dismissed):

PR #73, opened 2023-04-24
Feature/event handler

mattdean-digicatapult:
Event handler implementation and unit tests also includes some event processor implementation but this is not yet tested


src/lib/indexer/eventProcessor.ts
```diff
@@ -9,14 +13,104 @@ const processNameSet: Set<string> = new Set(processNames)
...
+      throw new Error(`Incompatible version ${version} for demand-create process`)
+    }
+
+    const newDemandId = outputs[0].id
+    const newDemand = outputs[0]
+
+    if (transaction) {
+      const id = transaction.localId
+      return {
+        demands: new Map([[id, { type: 'update', id, state: 'created', latest_token_id: newDemandId }]]),
```
jonmattgray, line 44: update `original_token_id` and `updated_at` as well?

src/lib/db/migrations/20230420123322_attachment-filename-nullable.ts
```diff
@@ -0,0 +1,13 @@
+import { Knex } from 'knex'
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.alterTable('attachment', (def) => {
+    def.string('filename').nullable().alter()
```
jonmattgray, line 5: db docs to update

src/lib/db/index.ts
```diff
@@ -52,21 +53,31 @@ const transactionColumns = [
   'updated_at AS updatedAt',
 ]
 
+export interface Transaction {
```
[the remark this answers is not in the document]
mattdean-digicatapult: I think `TransactionResponse` is missing the `localId` of the token refered to on the transaction. I think it's kind of important conceptually to separate the types as they serve different purposes (`TransactionResponse` is really for tsoa)

src/lib/indexer/eventProcessor.ts
mattdean-digicatapult: ok this doesn't happen automatically so I'll add it. I'll likely do it in the db module so it's always done correctly

mattdean-digicatapult:
Converting to draft as there are still tests failing

jonmattgray (approved):

PR #78, opened 2023-05-02
Feature/major version

mattdean-digicatapult:
Add major verison `v1` to all existing APIs. Note this doesn't change the location of `/swagger` or `/api-doc` as it is assumed an instance might support multiple major versions at once


src/controllers/v1/health/index.ts
```diff
@@ -2,9 +2,9 @@ import { Controller, Get, Route, Request } from 'tsoa'
 import { Logger } from 'pino'
 
 import * as Ex from 'express'
-import type { Health } from '../../models'
-import { logger } from '../../lib/logger'
-@Route('health')
+import type { Health } from '../../../models'
+import { logger } from '../../../lib/logger'
+@Route('v1/health')
```
jonmattgray, line 7: previously we haven't prefixed the `/health` endpoint with version e.g. in `dscp-identity-service` https://github.com/digicatapult/dscp-identity-service/blob/89e616ed8cb526412bb145f27786646df11cf59e/app/server.js#L51

just double checking we want to move health as well?

mattdean-digicatapult: agreed offline to leave the health endpoint at `/health`

n3op2 (approved):
nice one! quite a few updates

tsoa.json
```diff
@@ -20,7 +20,7 @@
     "specMerging": "recursive",
     "spec": {
       "paths": {
-        "/attachment": {
+        "/v1/attachment": {
```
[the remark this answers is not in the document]
mattdean-digicatapult: This was introduced a while back to fix the `accept` behaviour of the attachment endpoints. Bascially these entries override the spec declared in the handler as they can't be represented there

jonmattgray (approved):

PR #79, opened 2023-05-02
Use chain address if identity alias is null

jonmattgray:
Includes a regression test to assert that chain address gets used if (mocked) identity service responds with `alias: null`


jonmattgray:
Closed in favour of making the changes in `dscp-identity-service` instead

jonmattgray:
https://github.com/digicatapult/dscp-identity-service/pull/69

PR #80, opened 2023-05-03
3 persona docker

jonmattgray:
Adds a 3 persona docker compose for `MemberA`, `MemberB` and `Optimiser`

Each persona has a set of services:
- dscp-matchmaker-api (+ PostgreSQL)
- dscp-identity-service (+ PostgreSQL)
- dscp-node

There is a single `ipfs` instance

Process flows are added by `MemberA` during startup of their `dscp-matchmaker-api`


Dockerfile
```diff
@@ -18,10 +18,15 @@ FROM node:lts-alpine as service
 
 WORKDIR /dscp-matchmaker-api
 
+RUN apk add --update coreutils
```
jonmattgray, line 21: without this `dscp-process-management` fails because it uses `#!/usr/bin/env -S node --no-warnings`

mattdean-digicatapult (approved):
lgtm

dblane-digicatapult (approved):

PR #83, opened 2023-05-05
Feature/filter updated since

mattdean-digicatapult:
Add filtering on updated since. Also fixes various type issues and exposes created_at and updated_At fields


src/controllers/v1/transaction/index.ts
```diff
@@ -21,17 +22,26 @@ export class TransactionController extends Controller {
...
+   * @summary List transactions
    * @Query apiType lists all transactions by that type
    */
   @Response<BadRequest>(400, 'Request was invalid')
   @Response<NotFound>(404, 'Item not found')
   @Get('/')
   public async getAllTransactions(
     @Query() apiType?: TransactionApiType,
-    @Query() status?: TransactionState
+    @Query() status?: TransactionState,
```
n3op2, line 33: why not transaction status? not blocking as it was already implemented like that

src/lib/db/index.ts
```diff
@@ -95,10 +137,16 @@ export default class Database {
         [name]: () => this.client(name),
         ...acc,
       }
-    }, {}) as Models<() => Query>
+    }, {}) as Models<() => QueryBuilder>
```
n3op2, line 140: 🤦 used a wrong type

test/seeds/index.ts
```diff
@@ -22,7 +22,7 @@ export const seededAcceptTransactionId = 'd8eb8a94-222b-4481-b315-1dcbf2e07079'
 export const seededDemandAId = 'ae350c28-f696-4e95-8467-d00507dfcc39'
 
 export const seededMatch2Id = 'f960e4a1-6182-4dd3-8ac2-6f3fad995551'
-export const exampleDate = '2023-03-24T10:40:47.317Z'
+export const exampleDate = '2023-01-01T00:00:00.000Z'
```
n3op2, line 25: in some cases I like to use `mock-date` module so date kinda can be now

n3op2 (approved):
nothing that caught my eye, might be a good idea to get another set of eyes due to me being a little outdated in regards to demands

test/mocharc.json
```diff
@@ -1,4 +1,4 @@
 {
-  "timeout": 5000,
+  "timeout": 5000000,
```
jonmattgray, line 2: is this deliberate?

src/controllers/v1/transaction/index.ts
mattdean-digicatapult: We're already on the `/transaction` path so the word would be redundent I think

src/lib/db/index.ts
mattdean-digicatapult: No more I just needed to do some renaming so I could use the `Query` from `tsoa`. I felt this was a better name regardless

test/seeds/index.ts
mattdean-digicatapult: yeah and we use `sinon` so I could use `fakeTimers`. This was a smaller change to make it more obvious this was a mock value

jonmattgray (approved):

PR #89, opened 2023-05-10
Add `pending` state for entities not yet on chain

jonmattgray:
Add `pending` state for `demand` and `match2` to differentiate between an entity that's only been created locally (`pending`) and an entity that's been put on chain (`created`/`proposed`)


mattdean-digicatapult (approved):
lgtm

PR #142, opened 2023-09-01
Feature/l3 100

Ellenn-A:
Hi Guys, 

Could you pls have a look on the changes I have made? 
I have separated the seed data into separate files that tests use. 

Also sth odd is going on with the linting on the files I have added pls let me know how to fix that. 

For chain.test.ts there was the seed file being run before the test, but I have removed it as we don't need any seed data for that test. I replaced it with cleanup() but I'm not sure if we need that either. 

I have left the index file (original seed file) in, but it's not used anywhere anymore. Would you want me to remove it? 

also do I want to keep version as 1.2.43?


docker-compose-3-persona.yml
```diff
@@ -77,6 +77,7 @@ services:
       --alice
       --unsafe-ws-external
       --unsafe-rpc-external
+      --ws-max-connections 512
```
dblane-digicatapult, line 80: The additional `ws-max-connections` argument is not needed in `docker-compose-3-persona.yml` as we do not run automated tests against this file.

test/seeds/index.ts
```diff
@@ -1,493 +1,493 @@
...
+
+// const db = new Database().db()
+
+// export const cleanup = async () => {
+//   await db.attachment().del()
+//   await db.demand().del()
+//   await db.transaction().del()
+//   await db.match2().del()
+//   await db.demand_comment().del()
+// }
```
dblane-digicatapult, line 12: Delete file.

dblane-digicatapult (changes requested):
Looks good.  Try running the linter with the fix argument to fix the linting.
Delete the old seed index file.

mattdean-digicatapult (commented):
looks like a big improvement to me. Happy once @dblane-digicatapult is happy

dblane-digicatapult (approved):

PR #146, opened 2023-09-05
added identityStatus files

Ellenn-A:
Hello, 
please have a look and let me know how I should proceed. 
1. for identity.ts - can I move remaining methods to the IdentityClass? - if i do so, should I move URL_PREFIX to the constructor so that I always pass host and port every time a new instance of the class is made(if so, is it correct that I will always instatiate it with env.IDENTITY_SERVICE_HOST and env.IDENTITY_SERVICE_PORT - if so then I don't need to take in any parameters anyway and just have it as a private variable)? (I am asking because ipfs.ts and chainNode.ts are structured this way) (I have a question about how the environmental variables work in the 3-person yml - but having a call would be better) 
2. in identity.ts you'll notice I don't know how to get the identity version - any tips will be appreciated 
3. in healthHelper.ts - would I want to add a case for identityDown? 

4. question off-topic but are tests run against the 3-ppl yml? or just against the docker-compose.yml? - because it looks like some of my tests pass against one and some agains the other (once again if someone is willing to jump on a call, it'd be appreciated :))

### DIFF
--- a/src/lib/services/identity.ts
+++ b/src/lib/services/identity.ts
@@ -2,7 +2,7 @@ import { NotFound, HttpResponse } from '../error-handler'
 import env from '../../env'
 import { Status, serviceState } from '../service-watcher/statusPoll'
 
-const URL_PREFIX = `http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}/v1`
+const URL_PREFIX = `http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}`
 
 export default class IdentityClass {
   constructor() {}
@@ -11,7 +11,7 @@ export default class IdentityClass {
     try {
       const res = await getHealth()
       if (res) {
-        if (res.version.length < 1) {
+        if (!res.version.match(/\d+.\d+.\d+/)) {
           return {
             status: serviceState.DOWN,
             detail: {
@@ -39,7 +39,7 @@ export default class IdentityClass {
 }
 // TODO : refactor into the class above. Ticket: https://digicatapult.atlassian.net/browse/L3-182
 const getMemberByAlias = async (alias: string) => {
-  const res = await fetch(`${URL_PREFIX}/members/${encodeURIComponent(alias)}`)
+  const res = await fetch(`${URL_PREFIX}/v1/members/${encodeURIComponent(alias)}`)
 
   if (res.ok) {
     return await res.json()
@@ -53,7 +53,7 @@ const getMemberByAlias = async (alias: string) => {
 }
 
 const getHealth = async () => {
-  const res = await fetch(`http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}/health`)
+  const res = await fetch(`${URL_PREFIX}/health`)
 
   if (res.ok) {
     return await res.json()
@@ -63,7 +63,7 @@ const getHealth = async () => {
 }
 
 const getMemberBySelf = async () => {
-  const res = await fetch(`${URL_PREFIX}/self`)
+  const res = await fetch(`${URL_PREFIX}/v1/self`)
 
   if (res.ok) {
     return await res.json()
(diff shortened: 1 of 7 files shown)


src/lib/services/identity.ts
```diff
@@ -1,8 +1,59 @@
...
+}
 
 const URL_PREFIX = `http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}/v1`
 
+export default class IdentityClass {
+  constructor() {}
+
+  getStatus = async (): Promise<Status> => {
+    try {
+      const res = await getMemberBySelf()
```
mattdean-digicatapult, line 22: There should be a `/health` endpoint on the identity service that surfaces information about the identity instance. That should be better than using the self endpoint

src/lib/services/identity.ts
```diff
@@ -1,8 +1,59 @@
 import { NotFound, HttpResponse } from '../error-handler'
 import env from '../../env'
+export const serviceState = {
+  UP: 'up',
+  DOWN: 'down',
+  ERROR: 'error',
+} as const
+type serviceStateType = typeof serviceState
+export type SERVICE_STATE = serviceStateType['UP' | 'DOWN' | 'ERROR']
```
n3op2, line 9: since it's a type and not a constant why not 

```ts
export type ServiceState = 'UP' | 'DOWN' | 'ERROR'
```

src/lib/services/identity.ts
```diff
@@ -1,8 +1,59 @@
...
+          }
+        }
+        return {
+          status: serviceState.UP,
+          detail: {
+            version: 'some version',
+          },
+        }
+      }
+      return {
```
n3op2, line 40: we could just throw here? feels like a duplicate for line 47

test/integration/offchain/healthcheck.test.ts
```diff
@@ -15,6 +15,9 @@ const getSpecVersion = (actualResult: any) => {
 const getIpfsVersion = (actualResult: any) => {
   return actualResult?._body?.details?.ipfs?.detail?.version
 }
+const getIdentityVersion = (actualResult: any) => {
```
n3op2, line 18: not sure if version is being disclosed, we could maybe get from package.json since we are using as external lib (i mean identity service)

n3op2 (commented):
a few comments

src/lib/services/identity.ts
```diff
@@ -1,8 +1,59 @@
 import { NotFound, HttpResponse } from '../error-handler'
 import env from '../../env'
+export const serviceState = {
```
n3op2, line 3: since it's a constant why not label them like

```ts
export type ServiceState = 'up' | 'down' | 'unknown'.... // ignore names
export const SERVICE_UP: ServiceState = 'up' 
```

n3op2: well, I would keep types separate from constants

src/lib/services/identity.ts
```diff
@@ -1,8 +1,43 @@
...
 const URL_PREFIX = `http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}/v1`
 
+export default class IdentityClass {
+  constructor() {}
+
+  getStatus = async (): Promise<Status> => {
+    try {
+      const res = await getHealth()
+      if (res) {
+        if (res.version.length < 1) {
```
mattdean-digicatapult, line 14: This is ok but what might be better is a semver regex check that it is a valid version?

src/lib/services/identity.ts
```diff
@@ -17,6 +52,16 @@ const getMemberByAlias = async (alias: string) => {
   throw new HttpResponse({})
 }
 
+const getHealth = async () => {
+  const res = await fetch(`http://${env.IDENTITY_SERVICE_HOST}:${env.IDENTITY_SERVICE_PORT}/health`)
```
mattdean-digicatapult, line 56: I'd recommend refactoring the way `URL_PREFIX` is constructed to depuplicate logic here. At some point we'll likely enable https and it would be nice to only need to change this in one place

mattdean-digicatapult (changes requested):

mattdean-digicatapult (approved):
lgtm

n3op2 (approved):

PR #148, opened 2023-09-11
IdentityClass refactor

Ellenn-A:
Hello, 

I have added methods from outside of the IdentityClass into it and updated how methods are used in other files. 
Please let me know if you have any comments.

### DIFF
--- a/src/controllers/v1/_common/demand.ts
+++ b/src/controllers/v1/_common/demand.ts
@@ -17,22 +17,17 @@ import { demandCommentCreate, demandCreate } from '../../../lib/payload'
 import ChainNode from '../../../lib/chainNode'
 import env from '../../../env'
 import { parseDateParam } from '../../../lib/utils/queryParams'
-import { injectable } from 'tsyringe'
 import Identity from '../../../lib/services/identity'
 
-@injectable()
 export class DemandController extends Controller {
   demandType: 'demandA' | 'demandB'
   dbDemandSubtype: 'demand_a' | 'demand_b'
   log: Logger
   db: Database
   node: ChainNode
-  // identity: Identity
+  private identity: Identity
 
-  constructor(
-    demandType: 'demandA' | 'demandB',
-    private identity: Identity
-  ) {
+  constructor(demandType: 'demandA' | 'demandB', identity: Identity) {
     super()
     this.demandType = demandType
     this.dbDemandSubtype = demandType === 'demandA' ? 'demand_a' : 'demand_b'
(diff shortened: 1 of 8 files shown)


src/controllers/v1/_common/demand.ts
```diff
@@ -12,13 +12,15 @@ import {
...
 import { BadRequest, NotFound } from '../../../lib/error-handler/index'
-import { getMemberByAddress, getMemberBySelf } from '../../../lib/services/identity'
+import IdentityClass from '../../../lib/services/identity'
 import { TransactionResponse, TransactionType } from '../../../models/transaction'
 import { demandCommentCreate, demandCreate } from '../../../lib/payload'
 import ChainNode from '../../../lib/chainNode'
 import env from '../../../env'
 import { parseDateParam } from '../../../lib/utils/queryParams'
 
+const identityClass = new IdentityClass()
```
mattdean-digicatapult, line 22: We shouldn't be building this here. Ideally we should be instantiating the instance of IdentityClass (I dislike this name btw but it's ok) and passing this into the constructor of the controller. Look at https://github.com/digicatapult/dscp-matchmaker-api/blob/main/src/controllers/health/index.ts and https://github.com/digicatapult/dscp-matchmaker-api/blob/main/src/lib/service-watcher/index.ts for an example. I'm not sure exactly how parameters are passed to the singleton constructor; it might be best to look at documentation in https://github.com/microsoft/tsyringe and https://tsoa-community.github.io/docs/di.html#tsyringe

src/controllers/v1/_common/demand.ts
```diff
@@ -38,6 +40,7 @@ export class DemandController extends Controller {
       logger,
       userUri: env.USER_URI,
     })
+    this.identity = container.resolve(Identity)
```
mattdean-digicatapult, line 43: is there a reason to not use `@injectable`?

Ellenn-A: so I am trying it now I just pushed it. But when I use `@injectable` I have to pass it in the constructor like so: `
  constructor(private identity: Identity)` and then I din't know how to pass it e.g. in `v1/demandA`

mattdean-digicatapult: So if you pass it via `constructor(private identity: Identity)` the dependency injection framework should inject it for us. You won't have to instantiate it ourselves

mattdean-digicatapult: > That would make sense, but it doesn't do that - or at least not the way I made it Screenshot 2023-09-11 at 15 47 23

Ok I see. This is because this is a helper class for the two demand controllers (under `demandA`/`demandB`). In this instance the `DemandController` class shouldn't be the `injectable` but the `DemandAController` and `DemandBController` should be. They can then pass the `identity` instance passed to them to the `DemandController`. Does that make sense?

mattdean-digicatapult (approved):
looks good to me

PR #150, opened 2023-09-13
match2-cancel in processFlows

Ellenn-A:
I have included the match2-cancel in processFlows and I'm adding the restrictions for who can trigger the process now - Please have a look from line `2008` where I have attempted to add restrictions showing that only `MemberA` or `MemberB` can trigger the Match2 moving to cancelled state. 
- I have a question how to include this restriction on `DemandA` and `DemandB` - would we want to ad d a restriction that either of the members can move the demands to the cancelled state?


processFlows.json
```diff
@@ -1546,5 +1546,532 @@
...
+        "restriction": {
+          "SenderHasOutputRole": {
+            "index": 2,
+            "roleKey": "MemberB"
+          }
+        }
+      },
+      { "op": "And" },
+      { "op": "Or" },
+      { "op": "And" }
```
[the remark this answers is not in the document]
Ellenn-A: how do i specify what "or" relates to/encompasses ?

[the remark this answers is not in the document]
mattdean-digicatapult: It's a stack based evaluation. Each time we mention a `restriction` we push the value of that restriction onto the stack. Each `op` pulls two values of the stack. So lets say we have something like

```js
[
  RestrictionA,
  RestrictionB,
  RestrictionC,
  Or,
  And
]
```

The evaluation will then go like:

1. `stack=[RestrictionA]`
2. `stack=[RestrictionA, RestrictionB]`
3. `stack=[RestrictionA, RestrictionB, RestrictionC]`
4. `stack=[RestricitonA, (RestrictionB || RestrictionC)]`
5. `stack=[RestricitonA && (RestrictionB || RestrictionC)]` 

So at the end the result is `RestricitonA && (RestrictionB || RestrictionC)`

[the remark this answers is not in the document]
mattdean-digicatapult: So you're close but not quite there. Have another look and if you're still stuck we can talk it through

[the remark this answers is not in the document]
Ellenn-A: So I think it will be like this: 
```
    {
        "restriction": {
          "FixedInputMetadataValue": {
            "index": 2,
            "metadataKey": "state",
            "metadataValue": {
              "Literal": "acceptedFinal"
            }
          }
        }
      },
      { "op": "And" },
      {
        "restriction": {
          "FixedOutputMetadataValue": {
            "index": 2,
            "metadataKey": "state",
            "metadataValue": {
              "Literal": "cancelled"
            }
          }
        }
      },
      { "op": "And" },
      {
        "restriction": {
          "SenderHasOutputRole": {
            "index": 2,
            "roleKey": "MemberA"
          }
        }
      },
      {
        "restriction": {
          "SenderHasOutputRole": {
            "index": 2,
            "roleKey": "MemberB"
          }
        }
      },
      { "op": "Or" },
      { "op": "And" } 
```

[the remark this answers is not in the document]
mattdean-digicatapult: Yes that looks correct to me

Ellenn-A:
I have made the change as discussed above. 
My other issue is about cancelling the demands - I presume we also want to restrict who can do that. Please have a look at `1708-1728` and `1868-1887` where I have tried to include such restrictions for the demands.

processFlows.json
```diff
@@ -1705,7 +1705,27 @@
...
+        "restriction": {
+          "SenderHasOutputRole": {
+            "index": 0,
+            "roleKey": "MemberB"
+          }
+        }
+      },
+      { "op": "Or" },
+      { "op": "And" } ,
+
```
mattdean-digicatapult, line 1728: So these roles don't exist on demand tokens and I don't think this is logically necessary. We guarentee that the `match2` is attached to these tokens. In an earlier flow we make guarentees about the owners of those to the members on the match.

In otherwords I think these checks are unnecessary.

Ellenn-A: OK, I'll remove them

processFlows.json
```diff
@@ -1546,5 +1546,513 @@
...
+          "SenderHasOutputRole": {
+            "index": 2,
+            "roleKey": "MemberB"
+          }
+        }
+      },
+      { "op": "Or" },
+      { "op": "And" },
+
+
```
jonmattgray, line 2051: some weird formatting going on here with blank lines?

processFlows.json
```diff
@@ -1546,5 +1546,513 @@
...
+      { "op": "And" },
+      { "op": "Or" },
+      { "op": "And" }
```
jonmattgray, line 2054: is the `comment` metadata file required or optional? If it's required can restrict with:
```
        "restriction": {
          "FixedOutputMetadataValueType": {
            "index": 2,
            "metadataKey": "comment",
            "metadataValueType": "File"
          }
        }
   ```

Comment (Ellenn-A) [reply processFlows.json:2054]: I believe it's required. Thanks for pointing this out.

File: processFlows.json [hunk 5]
```diff
@@ -1546,5 +1546,519 @@
...
+          "FixedOutputMetadataValueType": {
+            "index": 2,
+            "metadataKey": "comment",
+            "metadataValueType": "File"
+          }
+        }
+      },
+      { "op": "And" },
+      { "op": "Or" },
+      { "op": "And" }
```
mattdean-digicatapult, line 2061: The last 2 op entries here are likely invalid. Have you tried running this into the node? You can do that with:

```sh
docker compose up -d
```

and then

```sh
npm run flows
```

Ellenn-A: Screenshot 2023-09-14 at 09 13 44
That worked - I was actually going to ask about the last two lines this morning.

mattdean-digicatapult (approved):

PR #152, opened 2023-09-14
Migration/l3 199

n3op2:
# What

changes: updating two tables
- match2 new state - cancellation
- transaction - new transaction type - cancelled


src/lib/db/migrations/20230914131825_math2-cancellation.ts
Ellenn-A, line 1: match :)

Ellenn-A (approved):

PR #159, opened 2023-09-18
List all cancellation transactions

Ellenn-A:
# List cancellation transactions
- added a route which lists cancellation transactions 
- added tests to check validity of new route
- addition to seed data to aid with the above mentioned testing 
- version bump (adjustment)

### DIFF
--- a/src/controllers/v1/match2/index.ts
+++ b/src/controllers/v1/match2/index.ts
@@ -390,6 +390,32 @@ export class Match2Controller extends Controller {
 
     return await this.db.getTransactionsByLocalId(query)
   }
+
+  /**
+   * @summary Get all of a match2's cancellation transactions
+   * @param match2Id The match2's identifier
+   */
+  @Response<NotFound>(404, 'Item not found.')
+  @SuccessResponse('200')
+  @Get('{match2Id}/cancellation')
+  public async getMatch2Cancellations(
+    @Path() match2Id: UUID,
+    @Query() updated_since?: DATE
+  ): Promise<TransactionResponse[]> {
+    const query: {
+      localId: UUID
+      transactionType: TransactionType
+      updatedSince?: Date
+    } = { localId: match2Id, transactionType: 'cancellation' }
+    if (updated_since) {
+      query.updatedSince = parseDateParam(updated_since)
+    }
+
+    const [match2] = await this.db.getMatch2(match2Id)
+    if (!match2) throw new NotFound('match2')
+
+    return await this.db.getTransactionsByLocalId(query)
+  }
 }
 
 const responseWithAliases = async (match2: Match2Row, identity: Identity): Promise<Match2Response> => {
--- a/src/controllers/v1/match2/index.ts
+++ b/src/controllers/v1/match2/index.ts
@@ -346,6 +346,32 @@ export class Match2Controller extends Controller {
   }
 
   /**
+   * @summary Get all of a match2's cancellation transactions
+   * @param match2Id The match2's identifier
+   */
+  @Response<NotFound>(404, 'Item not found.')
+  @SuccessResponse('200')
+  @Get('{match2Id}/cancellation')
+  public async getMatch2Cancellations(
+    @Path() match2Id: UUID,
+    @Query() updated_since?: DATE
+  ): Promise<TransactionResponse[]> {
+    const query: {
+      localId: UUID
+      transactionType: TransactionType
+      updatedSince?: Date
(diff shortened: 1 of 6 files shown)


src/controllers/v1/match2/index.ts
```diff
@@ -345,6 +345,25 @@ export class Match2Controller extends Controller {
     return transaction
   }
 
+  /**
+   * @summary Get all of a match2's cancellation transactions
+   * @param match2Id The match2's identifier
```
n3op2, line 350: you can also specify a type in curly brackets e.g. `{String}` which would get highlighted, since it's UUID, maybe we would like to stick to the UUID format. In that case it would be nice to have a tests as well for valid input

test/integration/offchain/match2.test.ts
```diff
@@ -217,6 +220,31 @@ describe('match2', () => {
...
+
+      expect(status).to.equal(200)
+      expect(body).to.be.an('array')
+      expect(body.find(({ id }: { id: string }) => id === seededMatch2CancellationId)).to.deep.equal({
+        id: seededMatch2CancellationId,
+        apiType: 'match2',
+        transactionType: 'cancellation',
+        localId: seededMatch2Id,
+        state: 'submitted',
+        submittedAt: exampleDate,
```
n3op2, line 235: for example date we can use `mockdate` lib https://www.npmjs.com/package/mockdate so no need to do hardcore prefixes/hardcoding

test/seeds/offchainSeeds/offchain.match2.seed.ts
```diff
@@ -39,6 +39,8 @@ export const seededDemandAAlreadyAllocated = '807d1184-9670-4fb0-bb33-28582e5467
 export const seededMatch2WithAllocatedDemands = '27965a5f-f3dd-4110-82e7-68f59bb02c2e'
 export const seededMatch2AcceptedA = '347411d3-3750-49dd-a548-88e9f2616d9c'
 export const seededMatch2AcceptedFinal = '85a50fd9-f20f-4a61-a7e4-3ad49b7c3f21'
+export const seededMatch2CancellationId = '85a50fd9-f20f-4a61-a7e4-3ad49b7c3f22'
+export const seededMatch2CancellationId2 = '85a50fd9-f20f-4a61-a7e4-3ad49b7c3f23'
```
n3op2, line 43:
```ts
const seededCancellations = ['uuid-id1', 'uuid-id2']
```

test/seeds/offchainSeeds/offchain.match2.seed.ts
```diff
@@ -298,6 +300,32 @@ export const match2Seed = async () => {
...
+      api_type: 'match2',
+      transaction_type: 'cancellation',
+      local_id: seededMatch2Id,
+      state: 'submitted',
+      created_at: exampleDate,
+      updated_at: exampleDate,
+      hash: transactionHash,
+    },
+  ])
+
```
n3op2, line 328:
```ts
(async (n) => {
  return new Array(n).fill({}).map((el) => db.transaction().insert([
    {
      id: generateRandomUUID(),
      api_type: 'match2',
      transaction_type: 'cancellation',
      local_id: generateLocalId(),
      state: 'submitted',
      hash: transactionHash,
    },
})()
```

mm... since we are starting to repeat

n3op2 (approved):
A few suggestions/comments

src/controllers/v1/match2/index.ts
Ellenn-A: doe this relate to the comment on 349-350?

test/seeds/offchainSeeds/offchain.match2.seed.ts
Ellenn-A: Provided I add this, How will I use it later on in the tests when I am trying to retrieve specific records by referring to them by their specific UUIDS -> should I put the newly generated UUIDs in an array - import this array to my test file and just pull uuids from there? -> But I presume that I cannot assert for the what the body contains then

test/seeds/offchainSeeds/offchain.match2.seed.ts
Ellenn-A: Only concern - does this potentially make the code less readable? - and another comment below

test/integration/offchain/match2.test.ts
Ellenn-A: Can I do this? I think the hardcoded date is important to know when we are doing the tests and we are filtering by records made before or after a certain date - or should I change the tests to use mockdate too?

PR #164, opened 2023-09-21
Bug/l3 215

mattdean-digicatapult:
This fixes a bug that appeared in https://github.com/digicatapult/dscp-matchmaker-api/pull/46. Essentially database transaction don't work because the `this` binding isn't changed in the `models` functions when recreating. This change refactors things to avoid the use of `this`


Ellenn-A (commented):
Looks good to me

n3op2 (approved):

PR #339, opened 2024-04-22
Swap out BearerAuth for OAuth2 in spec

mattdean-digicatapult:
Related to L3-230. This swaps out support for BearerAuth type authentication for Oauth2 now we're looking to use keycloak as an auth broker. Note this does not affect the way auth is handled by the application just how it is presented in the spec. The plan in time is to add scopes to each route so we can have a more fine grained approach

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@digicatapult/sqnc-matchmaker-api",
-  "version": "2.2.42",
+  "version": "2.3.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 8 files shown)


dblane-digicatapult (approved):

msterle (approved):
lgtm, although I'm surprised tsoa allows an empty flows object

mattdean-digicatapult:
> lgtm, although I'm surprised tsoa allows an empty `flows` definition in `securityDefinitions`

Yep I was surprised by that too. Even weirder the swagger UI is ok with it as well and that's the interface that primarily uses it

PR #348, opened 2024-05-01
Tmp/l3 null

n3op2:
# What

Updated authentication.js function to use jwks with some external libraries required for validation to happen.


mattdean-digicatapult:
Not needed after integrating https://github.com/digicatapult/tsoa-oauth-express

PR #461, opened 2024-09-18
Change logic of how we manage unprocessed blocks so these are recorde…

mattdean-digicatapult:
…d in the db

# Pull Request

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## PR Type

- [x] Bug Fix

## Linked tickets

SQNC-17

## High level description

Record unprocessed blocks in the db so if we OOM we don't lose progress

## Detailed description

SQNC-17 is caused fundementally by the fact that the list of unprocessed blocks is only tracked in memory. This change makes a new database table for unprocessed blocks so we can avoid this problem. Note this change is being done with urgency in mind and therefore:

* I've disabled unit tests around the updateUnprocessedBlocks logic. Onchain tests still pass and I've verified the change locally
* New unit tests will need to entirely change the test harness of the previous tests and therefore the old tests have little value
* New unit tests will follow in a separate PR

## Describe alternatives you've considered

None

## Operational impact

This change includes a db migration script and a fundemental change to how we process blocks. Given the situation with the l3 chain this is necessary but a bit scary. If the change fails we will need to create a new PR that undoes these changes. The rollback logic is already present. We may also need to truncate the `processed_blocks` table but there's no data of value in there anyway.

## Additional context

None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@digicatapult/sqnc-matchmaker-api",
-  "version": "3.0.80",
+  "version": "3.0.81",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@digicatapult/sqnc-matchmaker-api",
-      "version": "3.0.80",
+      "version": "3.0.81",
       "license": "Apache-2.0",
       "dependencies": {
         "@digicatapult/tsoa-oauth-express": "^0.1.42",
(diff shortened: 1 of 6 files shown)


jonmattgray (approved):

dblane-digicatapult (commented):
Scary but necessary

PR #521, opened 2024-11-20
improve healthcheck

Ellenn-A:
# Pull Request

## Checklist
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

## PR Type

- [ ] Bug Fix
- [ ] Chore
- [x] Feature
- [ ] Documentation Update
- [ ] Code style update (formatting, local variables)
- [ ] Breaking Change (fix or feature that would cause existing functionality to change)

## Linked tickets

SQNC-19

## High level description

Improve health check 

## Detailed description

Service is be marked as `DOWN` if the last processed block was more than new env `INDEXER_TIMEOUT_MS` which defaults to `30s`. (there are also scenarios for starting up time and processing/ catching up to past blocks)
In service watcher added a new `indexerStatus.ts` which uses new method from the `Indexer` to determine it's health based on timing of last processed blocks.


src/lib/indexer/index.ts
```diff
@@ -254,4 +273,41 @@ export default class Indexer {
...
+            detail: { message: 'Started more than 30s ago and no blocks have been processed or registered.' },
+          }
+        }
+      }
+    }
+    return {
+      status: serviceState.UP,
+      detail: { message: 'Service healthy.' },
+    }
+  }
```
mattdean-digicatapult, line 312: I think the logic here can be simplified to make it easier to reason about. It looks to me like there's a bug here if both times are null. 

Try to avoid `else` statements and instead rely on return statements in an `if` block. The order I think I'd consider is as follows:

1. if `startup` was less than `30s` ago then return `UP` 
2. if `lastProcessedBlockTime` and `lastUnprocessedBlockTime` are both `null` return `DOWN` (true because we know startup was longer than 30s ago)
3. consider the `latestActivityTime` as  `lastProcessedBlockTime || lastUnprocessedBlockTime`. (We know one must be truthy)
4. if `lastActivityTime` was less than `30s` ago return `UP`
5. return `DOWN` with the message derived from which one of the dates is truthy

mattdean-digicatapult (changes requested):
This looks better. I'd like to see if we can make the `getStatus` method clearer and also it would be good to see some tests

package.json
```diff
@@ -18,7 +18,7 @@
     "db:migrate": "npm run db:cmd -- migrate:latest --knexfile src/lib/db/knexfile.ts",
     "db:rollback": "npm run db:cmd -- migrate:rollback --knexfile src/lib/db/knexfile.ts",
     "coverage": "c8 npm run test",
-    "flows": "npx @digicatapult/sqnc-process-management@latest create -h localhost -p 9944 -u //Alice -f processFlows.json"
+    "flows": "npx @digicatapult/sqnc-process-management@2.2.131 create -h localhost -p 9944 -u //Alice -f processFlows.json"
```
mattdean-digicatapult, line 21: can you add it as a `devDependency` instead and then leave off the version here. We can then let renovate manage upgrades

src/lib/indexer/index.ts
```diff
@@ -6,31 +6,52 @@ import DefaultBlockHandler from './handleBlock.js'
...
+  constructor({ db, logger, node, handleBlock, retryDelay, startupTime, env }: IndexerCtorArgs) {
     this.logger = logger.child({ module: 'indexer' })
+    this.env = env
     this.db = db
     this.node = node
     this.gen = this.nextBlockProcessor()
     this.retryDelay = retryDelay || 1000
+    this.lastProcessedBlockTime = null
+    this.lastUnprocessedBlockTime = null
+    this.startupTime = startupTime
```
mattdean-digicatapult, line 54: can we anotate types on these please. At the moment I assume they're being infered as `any`. The first 2 should be `Date | null` the third just `Date`

src/lib/indexer/index.ts
```diff
@@ -254,4 +277,59 @@ export default class Indexer {
...
+        status: serviceState.DOWN,
+        detail: {
+          message: 'Last activity was more than 30s ago, no blocks were processed.',
+          startupTime: this.startupTime,
+          latestActivityTime: null,
+        },
+      }
+    }
+    const latestActivityTime = this.lastProcessedBlockTime || this.lastUnprocessedBlockTime
+    if (latestActivityTime === null) throw new Error('sth')
```
mattdean-digicatapult, line 305: you can remove this `if` by defining `latestActivityTime` before the `this.lastProcessedBlockTime === null && this.lastUnprocessedBlockTime === null` check. The equivalent is then if `latestActivityTime === null`

src/lib/indexer/index.ts
```diff
@@ -254,4 +277,59 @@ export default class Indexer {
...
+        },
+      }
+    }
+    return {
+      status: serviceState.DOWN,
+      detail: {
+        message: `Last activity was more than 30s ago. Last learned of block: ${this.lastUnprocessedBlockTime}`,
+        startupTime: this.startupTime,
+        latestActivityTime: latestActivityTime,
+      },
```
mattdean-digicatapult, line 332: as the only difference between these cases is the message I'd be tempted to combine them and then just set message using a ternary on `this.lastProcessedBlockTime`

src/lib/indexer/__tests__/index.test.ts
```diff
@@ -293,12 +297,52 @@ describe('Indexer', function () {
...
-      indexer = new Indexer({ db, node, logger, handleBlock })
+      indexer = new Indexer({ db, node, logger, handleBlock, startupTime, env })
       await indexer.start()
       const result = await indexer.processAllBlocks('1-hash')
 
       expect(result).to.equal(null)
       expect(handleBlock.called).to.equal(false)
     })
   })
+  describe('getStatus tests', function () {
```
mattdean-digicatapult, line 308: missing two tests for if we have an activity and that activity was more than 30s ago

src/lib/indexer/index.ts
mattdean-digicatapult, line 1: We need to make sure that once we exit `updateUnprocessedBlocks` `lastProcessedBlockTime` is not null. Otherwise we might continuously hit an error whereby we can't process the next block and this is an unhealthy state. I'd add after `await this.updateUnprocessedBlocks(lastProcessedBlock, lastKnownFinalised)` on linen ~ 120 a check that if `lastProcessedBlockTime` is null set it to the current date, probably with an appropriate comment

mattdean-digicatapult (changes requested):

src/lib/indexer/index.ts
Ellenn-A: so they already have these types:
```
  private startupTime: Date
  private lastProcessedBlockTime: Date | null // while we are running and up to date
  private lastUnprocessedBlockTime: Date | null // only for when we are catching up on unprocessed blocks
```

mattdean-digicatapult: yep my mistake

src/lib/indexer/index.ts
```diff
@@ -254,4 +278,52 @@ export default class Indexer {
...
+        status: serviceState.UP,
+        detail: {
+          message: 'Service healthy. Starting up.',
+          startupTime: this.startupTime,
+          latestActivityTime: currentDate,
+        },
+      }
+    }
+    const latestActivityTime = this.lastProcessedBlockTime || this.lastUnprocessedBlockTime
+    if (this.lastProcessedBlockTime === null && this.lastUnprocessedBlockTime === null) {
```
mattdean-digicatapult, line 296: this lint can be `if (latestActivityTime === null)`

src/lib/indexer/index.ts
```diff
@@ -254,4 +278,52 @@ export default class Indexer {
...
+      return {
+        status: serviceState.DOWN,
+        detail: {
+          message: 'Last activity was more than 30s ago, no blocks were processed.',
+          startupTime: this.startupTime,
+          latestActivityTime: null,
+        },
+      }
+    }
+    if (latestActivityTime === null) throw new Error('sth')
```
mattdean-digicatapult, line 306: then remove this line

src/lib/indexer/index.ts
```diff
@@ -97,11 +118,13 @@ export default class Indexer {
         }
 
         await this.updateUnprocessedBlocks(lastProcessedBlock, lastKnownFinalised)
+        if (this.lastProcessedBlockTime === null) this.lastProcessedBlockTime = new Date()
```
mattdean-digicatapult, line 121: defintely worth a comment here as to why we do this

src/lib/indexer/__tests__/index.test.ts
mattdean-digicatapult: I'm just counting `if` and `else` combinations. Even if I discount the redundent `if (latestActivityTime === null) throw new Error('sth')` line you have three `if` blocks that return and then the final return. There should be at least 4 tests threrefore (and as you're asserting message actually 5)

src/lib/indexer/index.ts
mattdean-digicatapult: I've been more explicit in further comments below

src/lib/indexer/index.ts
Ellenn-A: I don't think it can be...?
because this say if both are null at this point
`if (this.lastProcessedBlockTime === null && this.lastUnprocessedBlockTime === null)`
but this one says if either one of these is null at this point
`const latestActivityTime = this.lastProcessedBlockTime || this.lastUnprocessedBlockTime`

src/lib/indexer/__tests__/utils/helpers.ts
mattdean-digicatapult, line 1: is this now unused?

mattdean-digicatapult (approved):

PR #598, opened 2025-02-24
Feat/optionally use proxy

Ellenn-A:
# Pull Request

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

## PR Type

- [ ] Bug Fix
- [ ] Chore
- [x] Feature
- [ ] Documentation Update
- [ ] Code style update (formatting, local variables)
- [ ] Breaking Change (fix or feature that would cause existing functionality to change)

## Linked tickets

SQNC-99

## High level description

Update to optionally use proxy accounts to submit transactions.

## Detailed description

Added an env `PROXY`. If this is null we perform transactions etc. as Alice, however if `PROXY` is not null and there is a value for it, Alice is then Proxy for the persona, address we assign tot he `PROXY` env. 

So Alice would perform transactions on behalf of e.g. David.


src/env.ts
```diff
@@ -22,6 +26,7 @@ const env = envalid.cleanEnv(process.env, {
   NODE_PORT: envalid.port({ default: 9944 }),
   ENABLE_INDEXER: envalid.bool({ default: true }),
   USER_URI: envalid.str({ devDefault: '//Alice' }),
+  PROXY: nullableStr({ devDefault: '5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy' }), // needs to be nullable
```
mattdean-digicatapult, line 29: I think the semantics are confusing here. Can we rename to `PROXY_FOR`? Can you then also update the readme to make it explicit that this is the `ss58` address that transactions will be performed as

src/controllers/v1/_common/demand.ts
```diff
@@ -47,7 +48,17 @@ export class DemandController extends Controller {
...
+    // He is also the one who is associated with it in a db
+    let res: {
+      address: string
+      alias: string
+    }
+    if (env.PROXY !== null) {
+      res = self || (await this.identity.getMemberByAddress(env.PROXY, getAuthorization(req)))
+    } else {
+      res = self || (await this.identity.getMemberBySelf(getAuthorization(req)))
+    }
```
mattdean-digicatapult, line 61: This logic is repeated in several places. Can we move the entire of this to a new shared function that takes an `Identity` an `Env` and the authorization and does this logic and the relevant caching? That will allow you to deduplicate this logic and unit test it independantly

src/lib/chainNode.ts
```diff
@@ -150,7 +152,18 @@ export default class ChainNode {
     this.logger.debug('Preparing Transaction inputs: %j outputs: %j', inputs, fulfilledOutputs)
 
     await this.api.isReady
-    const extrinsic = this.api.tx.utxoNFT.runProcess(process, inputs, fulfilledOutputs)
+    //optionally use proxy here
+
+    let extrinsic
```
mattdean-digicatapult, line 157: This is an implicit any. I'm very surprised typescript isn't complaining. Please put a type on this. I also recommend removing the `else` clause below this by changing the logic to be more like:

```ts
let extrinsic: SomeType = this.api.tx.utxoNFT.runProcess(process, inputs, fulfilledOutputs)
if (this.proxyAddress) {
  extrinsinc =  this.api.tx.proxy.proxy({ id: this.proxyAddress }, null, extrinsic)
}
```

src/lib/payload.ts
```diff
@@ -20,22 +20,24 @@ export interface MetadataFile {
...
+          subtype: { type: 'LITERAL', value: demand.subtype },
+          parameters: { type: 'FILE', value: bs58ToHex(demand.ipfs_hash) },
+        },
       },
-    },
-  ],
-})
+    ],
+  }
+}
```
mattdean-digicatapult, line 40: I think this is just diff churn. Please revert

test/integration/onchain/demandA.test.ts
```diff
@@ -8,22 +8,29 @@ import { seed, cleanup, parametersAttachmentId, seededDemandAId } from '../../se
...
+import env from '../../../src/env.js'
+import { logger } from '../../../src/lib/logger.js'
 
 describe('on-chain', function () {
   this.timeout(80000)
   const db = new Database()
-  const node = container.resolve(ChainNode)
+  container.registerInstance(ExtendedChainNode, new ExtendedChainNode(logger, env))
+
+  const node = container.resolve(ExtendedChainNode)
```
mattdean-digicatapult, line 25: You shouldn't need to register this in each test suite. You should be able to just do `container.resolve(ExtendedChainNode)` and let `tsyringe` do the instantiation

test/integration/offchain/demand.test.ts
mattdean-digicatapult, line 1: Need tests for if not using a proxy. This applies to each suite. Note just one test for each transaction type is sufficient

test/integration/onchain/parallel-match2.test.ts
mattdean-digicatapult, line 1: I know it's beyond scope but github is warning about several `'transactionIds' is assigned a value but never used`. Can you look through and remove the assignments if these are unused

mattdean-digicatapult (changes requested):

test/integration/onchain/demandA.test.ts
Ellenn-A: so I tried doing this... but maybe I am creating the `ExtendedChainNode` in a wrong way because if I just try to resolve it I get an error: 
```
 Exception during run: Error: Cannot inject the dependency "logger" at position #0 of "ExtendedChainNode" constructor. Reason:
    TypeInfo not known for "Object"
```

mattdean-digicatapult: You need to look how it's done in `ChainNode` as it'll be similar. Something like

```ts
constructor(@inject(LoggerToken) logger: Logger, @inject(EnvToken) env: Env) {
  ...
}
```

Ellenn-A: so I did that originally, however that throws an error on the `@inject` that says `Decorators are not valid here.`

README.md
mattdean-digicatapult, line 1: can you update the docs on how to run the different test suites. They seem to be out of date

src/lib/chainNode.ts
```diff
@@ -150,7 +153,26 @@ export default class ChainNode {
...
+    // let extrinsic
+    // if (this.proxyAddress) {
+    //   extrinsic = this.api.tx.proxy.proxy(
+    //     { id: this.proxyAddress },
+    //     null,
+    //     this.api.tx.utxoNFT.runProcess(process, inputs, fulfilledOutputs)
+    //   )
+    // } else {
+    //   extrinsic = this.api.tx.utxoNFT.runProcess(process, inputs, fulfilledOutputs)
+    // }
```
mattdean-digicatapult, line 167: cleanup

src/utils/address.ts
```diff
@@ -0,0 +1,22 @@
+import type express from 'express'
+import { Env } from '../env.js'
+import Identity from '../lib/services/identity.js'
+import { getAuthorization } from '../lib/utils/shared.js'
+
+export const determineAddress = async (
```
mattdean-digicatapult, line 6: can I suggest we memoize the valud of `self` within this file rather than have separate caches elsewhere? That will remove `self` as a parameter and mean we're caching it more effectively.

I'd also rename this to `determineSelfAddress` to be more descriptive about what this is

test/integration/onchain/demandA.test.ts
mattdean-digicatapult: alright, we'll sort this out in https://digicatapult.atlassian.net/browse/SQNC-137

test/integration/onchain/demandA.test.ts
```diff
@@ -8,22 +8,29 @@ import { seed, cleanup, parametersAttachmentId, seededDemandAId } from '../../se
...
 import { withAppAndIndexer } from '../../helper/chainTest.js'
 import { container } from 'tsyringe'
 import { filterRejectedAndAcceptedPromises } from '../../helper/parallelTests.js'
+import { withProxy } from '../../helper/proxy.js'
+import ExtendedChainNode from '../../helper/testInstanceChainNode.js'
+import env from '../../../src/env.js'
+import { logger } from '../../../src/lib/logger.js'
 
-describe('on-chain', function () {
+describe('on-chain via proxy', function () {
```
mattdean-digicatapult, line 20: let's reduce the churn for when the non-proxy version goes away. Can we call these `on-chain` and then the new test suite `on-chain proxyless` or similar please

test/integration/proxyless/proxyless-demandA.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import Database from '../../../src/lib/db/index.js'
+import ChainNode from '../../../src/lib/chainNode.js'
+import { pollTransactionState, pollDemandState } from '../../helper/poll.js'
+import { withAppAndIndexer } from '../../helper/chainTest.js'
+import { container } from 'tsyringe'
+
+describe('on-chain', function () {
+  this.timeout(80000)
+  const db = new Database()
+  const node = container.resolve(ChainNode)
```
mattdean-digicatapult, line 19: can I recommend we have an `it` that resolves the env and asserts we're not using a proxy.

mattdean-digicatapult (changes requested):

